PR #14, opened 2025-01-17
Add Z-machine color support

dstelzer:
This pull request enables the `color` and `background-color` CSS properties on Z-machine. It's an enhancement to the compiler, but not a change to the language—this was always supported in the web interpreter, for example.


bkirwi:
I am very much not a C programmer, but overall this seems well-scoped and the code makes sense to me.

Two questions:
- One alternative would be to have separate routines for setting "true" and "standard" colours... IIUC nothing forces them to go through the same runtime routine. We could shift some of the math out of runtime and into the compiler, by statically determining which sort of colour we have and invoking the right logic for it instead of branching at runtime. (While I love moving work to compile time, this is only a very small amount, so may not be worth the effort when there's other stuff to do.)
- Could you say a bit about how resets work? For example, if I have a bunch of nested differently-coloured boxes, how do we ensure that we reset the colour to the outer colour once we leave the inner?

dstelzer:
Good questions!
- Definitely possible, since the compiler has separate logic for parsing "true" colors and "standard" colors. But since the decision needs to be made for both the foreground and the background color separately, I figured the extra runtime logic (a single comparison and subtraction each) was less ugly than putting four new routines into every game file (for each combination of foreground and background type).
- Currently, this is the done the same way as style resets: every time Dialog enters a new box or span, it pushes the old values of REG_FGCOLOR and REG_BGCOLOR onto the stack, and every time it exits a box or span, it pulls those values back off. Right now, this is even done with status areas, which isn't actually necessary—it's a fatal error to begin a status area inside a div or span—but in case that becomes possible, I wanted to future-proof it. (Plus it's easier to just always do the pushing and popping instead of having separate logic to bypass it; two words of stack space isn't that valuable. It's not done for REG_STYLE because that's handled inside the R_BEGIN_BOX routines, but Dialog's Z-machine assembler can't handle passing more than three arguments to a routine.)

heasm66 (approved):
I'll accept it to test new ruleset for main branch.

PR #16, opened 2025-01-20
Bump version to 0m04 (in development) + placeholder for it in readme.txt

heasm66:
As it say in the title.

VERSION is set in the Makefile as a compiler constant. The documentation also needs to be updated eventually.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -38,6 +38,11 @@ Project website:
 
 Release notes:
 
+	0m/04, Lib 0.46 (Manual revision 31):
+
+		Compiler: Enables the color and background-color CSS properties on 
+		Z-machine.
+
 	0m/03, Lib 0.46 (Manual revision 31):
 
 		Library: Changed how darkness is handled: Unlit objects are now
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,5 +1,5 @@
 CC		= gcc
-CFLAGS		+= -O3 -Wall -DVERSION=\"0m/03\"
+CFLAGS		+= -O3 -Wall -DVERSION=\"'0m/04 (in development)'\"
 WININCLUDE	= winglk
 WINLIB		= ../prebuilt/win32
 


bkirwi:
#17 also touches the version, but this seems like a fine place to discuss versioning in general. 

For the major version: [Linus](https://intfiction.org/t/dialog-wishlist/70278/140?u=bkirwi) suggests that we start versioning from `1a`. I assume everyone is okay with this?

For the minor version: on major version bumps the minor version seems to reset to 1, not the traditional 0. (So the first release with the new major version would be `1a/01`. On the linked branch @dstelzer suggests `1a/02`, since that's the second PR. However, [looking at the changelog](https://github.com/Dialog-IF/dialog/blob/main/readme.txt), Dialog has traditionally collected many changes in a single minor release. I have a mild preference for having individual PRs leave the version alone and then bumping it occasionally as part of an explicit release, but I am willing to be overruled on this!

For the `(in development)` annotation: I approve but don't feel strongly. Earlier I was going to propose adding "community edition" or something to the version string, but it sounds like Linus is comfortable with leaving the `1a` versions to the community project entirely, so that no longer feels strictly necessary...

dstelzer:
I read something like 1a02 as major version 1, minor version a, subminor version 2, so I would bump the subminor version whenever an observable change comes out—in this case, the Z-machine styling will be different after the new PR, so I think it deserves a new patch number.

bkirwi:
I'd definitely agree that it's a significant enough change to require a version bump when it's released! Is the idea that every merge to main counts as a release?

dstelzer:
That would be my preference. That way, anyone cloning the main branch gets the latest version.

bkirwi:
Got it!

I don't think I've ever seen a community project that used that approach. I feel like it would be harder to maintain the usual expectations that come with a "release", like features being complete and tested and documentation being up to date. It also seems out of step with how Dialog has been developed in the past, where you see often several unrelated language features in the same release. But I'm up to try it if there's consensus!

(I have contributed to a few projects that have a separate `release` and `development` branch, and make `release` the default in the Github UI. In my experience this is fiddlier than the default, since every PR requires some additional redirection, but I believe it helps with this specific problem.)

Anyways, enough from me - I'll let the other folks on the thread weigh in.

dstelzer:
Fair, but isn't the point of requiring reviews on pull requests that features should be complete and tested and documented?

heasm66:
The 1a part is pretty clear from Linus. I obviously like the method of stacking a couple of changes in the last part before bumbping it up, so there's a current version in development and a stable previous version. That said, I'm open to use @dstelzer's suggestion instead. 

What do we do with the 100th PR, tough? (Is there space enough in the z-machine header to write a longer version string?)

dstelzer:
There are eight bytes at the end of the header which have no official meaning; Infocom puts eight bytes of ASCII there (the player's username), Inform 6 puts "InfoX.XX" (e.g. Info6.12), Dialog leaves the first one blank and puts "DiaXXXX" (e.g. Dia1a02", slash not included) in the remaining seven.

That said, I'm not that invested in bumping the version number, so if you'd like it to be a separate process from making pull requests, that's fine too.

dstelzer:
Dammit, I tried to merge the zstyling branch without including the version bump, but I think I messed it up. Someone who has full write access to main, feel free to revert that.

dstelzer:
While we discuss this, we should also figure out versioning for the library and the documentation.

Neither of these have any coding-related requirements, the way the Z-machine header does, so we can really do anything we want. But right now, the library is version 0.46, and the manual is version 31.

I tentatively propose switching both of them to Major.Minor.Patch versioning, and bumping the major version to 1, as we did with the compiler. Whatever we decide to do with the patch numbers for the compiler, we would do the same with the others.

dstelzer:
Good point! I should have said major.minor.patch versioning, I think you're right that using a letter isn't a good idea for the other components.

Then we can set all the major versions to 1 for our version, in accordance with Linus's wishes, and go from there.

bkirwi:
How would y'all feel about using the same version for everything?

Up to this point, a single version of the documentation has documented precisely one compiler and one library version. If we're developing them in the same repo, and testing only the current versions against each other, I don't see a strong reason not to give them the same version number.

An alternative that we haven't discussed is "keep doing what Linus did", which I think is probably the default we should choose unless we have a decent reason otherwise. (But "three version numbers is confusing" is, to me, a decent reason.)

dstelzer:
The main argument against it (that I can see) is that it makes it harder to keep track of what components have changed. I'm expecting library changes to be much more common than compiler changes, for example, and I wouldn't want to make everyone rebuild the compiler from source every time that happens.

My vote is to set the compiler version to 1a01, the library version to 1.0.1, and the documentation version to 1.0.1, for the start of our new repository, and then bump the patch version for the appropriate component every time we update it. Having that be when a pull request is merged ensures that we always have an unambiguous way to know what version of the code someone was using when they report a bug.

But also, I haven't really dealt with the logistics of open-source projects before, so there may be problems with this approach I'm not seeing.

bkirwi:
> The main argument against it (that I can see) is that it makes it harder to keep track of what components have changed. I'm expecting library changes to be much more common than compiler changes, for example, and I wouldn't want to make everyone rebuild the compiler from source every time that happens.

That's fair! Personally I don't expect folks to want to update the library that often either... or at least I personally will probably not upgrade my projects more than a few times a year. (The other side, how annoying it is to upgrade the compiler, depends on...)

> Having that be when a pull request is merged ensures that we always have an unambiguous way to know what version of the code someone was using when they report a bug. But also, I haven't really dealt with the logistics of open-source projects before, so there may be problems with this approach I'm not seeing.

I'm nervous about treating our development branch as the release branch, for reasons such as:
- A PR breaks the build in a way that CI doesn't catch.
- A PR breaks the build in a way that CI _would_ have caught, but someone else merged a PR in the meantime that changed something else.
- Someone typos the version number.
- I correctly increment the version number in my PR, but some PRs have merged in the meantime so now it's wrong.
- We both correctly increment the version number and merge around the same time. Our changes _don't_ conflict, but now there's multiple commits in `main` with identical version numbers but different behaviour.
- An outside contributor submits some patches and doesn't understand which version numbers to bump.
- We want to merge a really big feature that requires sweeping changes. Either we land small PRs that leave `main` in an awkward intermediate state that we'd rather not expose to users, or we bundle everything in one big merge that is harder to review and increases the risk of some other issue.

I've come across a few projects that use the current commit hash as the version number, which avoids some of those problems. I'm not aware of any open-source project that bumps the version on every merge. Almost all the projects I've interacted with make periodic releases of known-good commits chosen from `main`.

bkirwi:
But also baked into that model is the idea that you point users to something like https://github.com/Dialog-IF/dialog/releases instead of having them build the current `main` from source. (Personally, I'd love to get things set up so that we pre-build binaries for common systems and share those out along with the current source snapshot.) If we're expecting most users to navigate here and then `git fetch` I can see how the balances would feel different!

dstelzer:
Fair! If you have ideas for a better way to distinguish between development and production versions, that makes sense to me. I'm just not sure, with how few people are involved in this project, if there's really going to be much more review than pull requests already get.

dstelzer:
Well, the fact that I just caused two major bugs in the main branch because they didn't show up in my test cases has convinced me that some more review is a good idea! If someone wants to take over the role of periodically testing, bumping the release number, and posting an official release, I think that's a fine idea.

bkirwi:
This was superseded by #51, based on discussion here and elsewhere. Thanks all!

PR #17, opened 2025-01-20
Allow style stacking on Z-machine

dstelzer:
Expands the Z-machine style system in four ways:
- The CSS property `text-decoration: reverse;` sets reverse-video style
- A single style class can now set multiple styles, like bold italics
- Styles can be inherited between spans and divs; a bold span inside an italic span will be bold italic, unless the inner span explicitly specifies not italic
- Styles work normally in the upper window; reverse-video is applied by default, but can be turned off with `text-decoration: none;` (though note that the upper window is always monospace and this can never be turned off per the Z-machine standard)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,8 @@
 *.o
+dgdebug
+dialogc
+*.dg
+*.z5
+*.z8
+*.zblorb
+*.aastory
--- a/src/Makefile
+++ b/src/Makefile
@@ -1,5 +1,5 @@
 CC		= gcc
-CFLAGS		+= -O3 -Wall -DVERSION=\"0m/03\"
+CFLAGS		+= -O3 -Wall -DVERSION=\"1a/02\"
 WININCLUDE	= winglk
 WINLIB		= ../prebuilt/win32
 
--- a/src/frontend.c
+++ b/src/frontend.c
@@ -2966,8 +2966,8 @@ int frontend(struct program *prg, int nfile, char **fname, dictmap_callback_t di
 				(bc->style & STYLE_BOLD)? "yes" :
 				(bc->unstyle & STYLE_BOLD)? "no" : "inherit");
 			printf("\tMonospace:\t%s\n",
-				(bc->style & STYLE_MONOSPACE)? "yes" :
-				(bc->unstyle | STYLE_MONOSPACE)? "no" : "inherit");
+				(bc->style & STYLE_FIXED)? "yes" :
+				(bc->unstyle | STYLE_FIXED)? "no" : "inherit");
 			if(bc->color < 0) { // Print negatives in decimal
 				printf("\tColor:\t%d\n", bc->color);
 			} else { // Positives in hex
--- a/src/frontend.c
+++ b/src/frontend.c
@@ -2910,16 +2910,16 @@ int frontend(struct program *prg, int nfile, char **fname, dictmap_callback_t di
 					}
 				} else if(1 == sscanf(str, "font-style : %s", param)) {
 					if(!strcmp(param, "italic") || !strcmp(param, "oblique")) {
-						bc->style = STYLE_ITALIC;
+						bc->style |= STYLE_ITALIC;
 					}
 				} else if(1 == sscanf(str, "font-weight : %s", param)) {
 					if(!strcmp(param, "bold")) {
-						bc->style = STYLE_BOLD;
+						bc->style |= STYLE_BOLD;
 					}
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -1 +1,8 @@
 *.o
+dgdebug
+dialogc
+*.dg
+*.z5
+*.z8
+*.zblorb
+*.aastory
```
bkirwi, line 8: :bowing_man:

src/Makefile
```diff
@@ -1,5 +1,5 @@
 CC		= gcc
-CFLAGS		+= -O3 -Wall -DVERSION=\"0m/03\"
+CFLAGS		+= -O3 -Wall -DVERSION=\"1a/02\"
```
bkirwi, line 2: Mentioned this in #16, which conflicts with this change. I'll leave discussion of the versioning approach to that branch!

src/frontend.c
```diff
@@ -2910,16 +2910,28 @@ int frontend(struct program *prg, int nfile, char **fname, dictmap_callback_t di
...
-						bc->style = STYLE_ITALIC;
+						bc->style |= STYLE_ITALIC;
+					}else if(!strcmp(param, "normal")) {
+						bc->unstyle |= STYLE_ITALIC;
 					}
 				} else if(1 == sscanf(str, "font-weight : %s", param)) {
 					if(!strcmp(param, "bold")) {
-						bc->style = STYLE_BOLD;
+						bc->style |= STYLE_BOLD;
+					}else if(!strcmp(param, "normal")) {
```
bkirwi, line 2920: Minor formatting issue:

```suggestion
					} else if(!strcmp(param, "normal")) {
```

Here and elsewhere!

src/frontend.c
```diff
@@ -2910,16 +2910,28 @@ int frontend(struct program *prg, int nfile, char **fname, dictmap_callback_t di
...
-					// %s stops at first whitespace, so use str:
+					// %s stops at first whitespace, so use str instead
 					if(strstr(str, "monospace")) {
-						bc->style = STYLE_FIXED;
+						bc->style |= STYLE_FIXED;
+					}else if(strcmp(param, "inherit")) { // Something that's not monospace was specified, and it was *not* inherit
+						bc->unstyle |= STYLE_FIXED;
+					}
+				} else if(1 == sscanf(str, "font-decoration : %s", param)) {
+					if(strstr(str, "reverse")) { // This is not a standard CSS property, but there is no standard CSS property for reverse video, and unrecognized property values are explicitly not an error in CSS
```
bkirwi, line 2931: Mild preference for `reverse-video`, if that works for you? (Less ambiguous and less likely to conflict with (the admittedly unlikely possibility of) some future version of CSS defining this differently.

src/runtime_z.c
```diff
@@ -3450,49 +3477,51 @@ struct rtroutine rtroutines[] = {
...
+			{Z_SPLIT_WINDOW, {VALUE(REG_LOCAL+0)}}, // And actually split the screen
 			{Z_SET_WINDOW, {SMALL(1)}},
 
-			{Z_CALL1S, {ROUTINE(R_GET_FULLWIDTH)}, REG_XFULLSIZE},
+			{Z_CALL1S, {ROUTINE(R_GET_FULLWIDTH)}, REG_XFULLSIZE}, // Update REG_XFULLSIZE
 
-			{Z_TEXTSTYLE, {SMALL(0)}},
-			{Z_TEXTSTYLE, {SMALL(1)}},
+	//		{Z_TEXTSTYLE, {SMALL(0)}}, // Turn off all text styles
+	//		{Z_TEXTSTYLE, {SMALL(1)}}, // Turn on reverse video style
```
bkirwi, line 3515: I'd be tempted to delete the commented-out code... we can always fish this back out from source control if we need to.

bkirwi (approved):
Nice! Once again the overall gist makes sense to me, though someone who has more C than me may want to take a look also.

I'd appreciate if we avoided merging the version bump until #16 gets hammered out, to avoid thrashing. But if you'd rather get this in we can always sort that out later.

My understanding is that this obviates the style predicates, but doesn't actually remove them yet... does that sound about right?

dstelzer:
Correct on obviating the style predicates but not removing them. Removing them doesn't really gain us anything over deprecating them, right now—the main benefit would be substantially simplifying the Å-machine's I/O architecture, and I haven't touched the Å-machine yet.

I have no problems with any of your suggested changes!

PR #18, opened 2025-01-20
Add Antora version of the manual

hlship:


dstelzer:
Can we integrate the Antora compilation into Github somehow, or provide compiled versions alongside the source in the repository? Node.JS is a pretty hefty dependency (several gigabytes!) to bring into the build process just for the documentation.

hlship:
Ultimately, I hope to set this up the same way I did with Pedestal; there's a separate repository that exists just for the Antora manual build, and it will rebuild after merges to main in the dialog repo.

That works well because there's a 1:M relationship between the manual repo and the dialog and potentially other repos that will contribute to the manual (I'm thinking dialog-tool and dialog-extensions in the short term, and more community things in the longer term).

But it does cause a problem including a manual with future dialog releases and if that stays a priority, an alternate approach may be needed.

hlship:
I do value the idea of having downloadable documentation with dialog, since dialog is so nicely self-contained, it should be possible to develop when there is no internet.  I don't have an immediate solution on how to accomplish that, however.

dstelzer:
I've been very impressed with how Github's automatic build tools have worked for the main tools, so if it's possible to build the manual on their servers the same way, I think that would be perfect.

Unfortunately, I don't know enough about Antora or about Github builds to say if that's possible. But this is looking great so far!

hlship:
For Pedestal, each merge to master in the main pedestal repo triggers an action in the pedestal-docs repo; it rebuilds the documentation (currently, two stable versions from maintenance branches, and one unstable version in master) and publishes.  The docs are always up-to-date within about a minute of a commit to master.

I believe I can replicate that for Dialog; the big question is how to publish the resulting docs to a static website, such as an organization-level github.io.  There's almost certainly a GitHub action for that, but I'll sort that out once I have the documentation converted.

dstelzer:
Wonderful! In that case, that's all my objections settled. I'm fully in favor.

hlship:
There's a chicken-and-the-egg problem for including the generated documentation as part of the release, as had been traditional.  What gets packaged may be a "very recent" version of the docs, and there will be developer guidelines about producing releases that includes a short wait for the manual to build first.

hlship:
There's a chicken-and-the-egg problem for including the generated documentation as part of the release, as had been traditional.  What gets packaged may be a "very recent" version of the docs, and there will be developer guidelines about producing releases that includes a short wait for the manual to build first.

dstelzer:
Congratulations!

hlship:
This content represents a complete translation of the existing manual.

I'm working on a second repository which will be responsible for building the actual HTML manual from these sources.

dstelzer (approved):
This all looks good to me! And having this in the main repo seems like a good thing for keeping compiler and documentation in sync. Approved.

hlship:
Yes, for example, to describe the new --numeric option to dgdebug.

hlship:
I'm still finding tiny issues, but those can be addressed in a subsequent PR.

bkirwi (approved):
I have not read literally all the docs, but I think this is a pretty clear jump forward for maintainable docs, so I'm +1 to merging.

One thought I wanted to mention up front...

I think the source docs are quite readable on their own, but I'd find it easier to find a specific doc to update if they included their section number as a prefix - eg. `11-miscfeat.adoc`. Definitely not a blocker, but if you really like the idea might as well do it now to make preserving history easier!

bkirwi:
> There's a chicken-and-the-egg problem for including the generated documentation as part of the release, as had been traditional.

I'd prefer not to actually check those into the repo, but I think they'd make total sense as part of a Github Release. (Along with prebuilt binaries etc.) OTOH maybe the fancy Antora feature of being able to display multiple documentation versions means it's less necessary to bundle them up together. Will check in on this as we approach a full release!

hlship:
Not sure I understand how having the chapter number be part of the file name is going to help ... and if we at some point split longer chapters into smaller parts, perhaps, that will be an impediment.  The module, "lang" vs. "lib", is now part of the canonical URL, which is more helpful.  I did keep all the original file names, though they are now split between multiple directories (each an Antora module within the Dialog  content source).

dstelzer:
Numbering the chapters seems like it might help keep the table of contents in sync, at least, so that would be a reason for it. I'm not wedded to the idea, though.

hlship:
I only kept the chapter numbers around because that's how Linus did it, and I'm trying to replicate the documentation content, form, and structure as closely as possible, at least initially.

bkirwi:
> Not sure I understand how having the chapter number being part of the file name is going to help [...]

Concretely, when reviewing the docs in a text editor, I found myself looking for things like "the first chapter of the language docs" and "this thing I remember being about halfway through the library docs" or "chapter three". Having the chapter numbers in the filename would make this sort of navigation quicker.

Anyways, if you're not compelled it's definitely not a blocker for me! Feel free to merge when ready.

hlship:
Who has write access and can actually merge?

dstelzer:
Oop! I thought you had permission to do that yourself.

For now, I'll merge it.

PR #19, opened 2025-01-21
Build via CI

bkirwi:
For now, this just checks that the source compiles. In the future I'd imagine also running tests etc., but it's a start!

IIUC, it's set to trigger on any PR and on a merge to `main`.


heasm66 (approved):
It's a beginning.

PR #21, opened 2025-01-21
Fix signed comparison bug in Z-machine runtime

dstelzer:
R_WORDMAP was comparing addresses to object numbers using the standard Z-machine JG opcode. Unfortunately, this opcode does signed comparison, so any addresses above 0x8000 would be treated as negative, and thus considered to be object numbers.

One line of code in the runtime fixed this, but I've also added various tracing instrumentation to the compiler in order to find the problem in the first place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dialogc
 *.z8
 *.zblorb
 *.aastory
+*.log
--- a/src/backend_z.c
+++ b/src/backend_z.c
@@ -809,7 +809,7 @@ void prepare_dictionary_z(struct program *prg) {
 	uint32_t uchar;
 	char chbuf[2];
 	struct word *w;
-	char runtime[7] = {8, 13, 32, 16, 17, 18, 19};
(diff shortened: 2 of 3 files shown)


bkirwi:
I won't have a chance to look at this for a day or two, but bravo! Great find.

dstelzer:
Thanks! The bug fix itself should be pretty uncontroversial—it's just one additional line of Z-machine assembly—but this branch also adds a bunch of instrumenting to the wordmap part of the Z-machine frontend, in order to find the bug in the first place.

It's locked behind verbosity 3, which already dumps so much detail that you're not going to use it without something like grep, so I don't really see the harm, but I also wouldn't be opposed to putting it behind an `#ifdef`, since I don't foresee it being especially useful beyond this specific instance.

bkirwi (approved):
It's great!

I'm happy to leave the verbose logging in. I'm happy to consider it not part of the stable interface, so we can change it freely later if we change our minds.

PR #27, opened 2025-01-24
More builds and testing for CI

bkirwi:
Includes:
- Windows builds. Turns out the distribution zips had `Glk.dll` precompiled; this repo contained only headers. I added that back in, along with license info etc.
- Tests. I've imported `regtest.py` and Linus' Cloak of Darkness, and added a simple walkthrough test for it.


bkirwi:
I don't think Linus released Cloak under any specific license, though I suspect in practice it's fine? If we're unsure... @dstelzer, I think you mentioned you might have something we can include and add a regtest for?

dstelzer:
My main test so far has been my regression tests for _Miss Gosling's Last Case_, but in practice that's not quite enough: it tests a lot of behavior in both the debugger and the Å-machine, but it didn't catch those crashes I fixed in the other branch, because _Miss Gosling_ doesn't use floating divs, and never uses an inline status bar on the Z-machine (it checks `(interpreter supports inline status bar)` and puts the info in the top status bar if not).

So I think using RemGlk for JSON will help, since it'll also let us check for regressions in things like the status bar that dumbfrotz doesn't find.

dstelzer:
I added the Miss Gosling tests I used to `/test/gosling`. There's a Makefile in there with details. It's a fairly extensive test of the system in its various respects.

However, I'm not sure how to integrate this with your testing framework, and I don't really want to mess with your system without understanding it. I think I may already have messed something up because git gave me an error about Glk.dll.

So for now, I'll just leave you with the Makefile, and let you do the integration. Hope that's all right!

dstelzer:
(I'm also not sure how to have diff not complain about the differing serial numbers. That's probably something regtest can do a better job at!)

bkirwi:
> However, I'm not sure how to integrate this with your testing framework, and I don't really want to mess with your system without understanding it.

I tried to stick to as vanilla of setup with Zarf's regtest.py as possible... though I've never used it before, it turns out to do more or less the right thing if you just paste a transcript in and tidy it up a bit.

> I think I may already have messed something up because git gave me an error about Glk.dll.

This seems to have just been a bad rebase - it's all fine now.

I've done what appears to be enough integration to get everything humming, if you want to take another look?

bkirwi:
We should give some thought to licensing... do you have a sense of what license you'd like to use for Gosling here?

dstelzer:
Thanks! That looks great so far. And argh, I seem to have somehow messed up Glk.dll when I pulled again, I can't figure out why my git keeps doing this. Is there something else I should be doing than `git pull` then `git checkout build`?

If you want to adapt the _Miss Gosling_ tests to use regtest.py, I'm not at all opposed, I just don't really know how to do it. But I can take a look next week if desired.

I'd also like to have a couple tests using RemGlk, to look for issues with the status bar and styling and all that, but I honestly don't understand how floating divs are supposed to work in Dialog. If someone can construct an appropriate test for that, I can add my color and styling tests from when I was working on that branch.

As far as licensing, I haven't actually given much thought to it before. My immediate instinct is Creative Commons Attribution, which is what I release most of my creative stuff under, but I don't know how well that works for software. Whatever license I end up using, though, I'm happy to dual-license or whatever to ensure it's compatible with the repository license as a whole.

bkirwi:
> I can't figure out why my git keeps doing this. Is there something else I should be doing than `git pull` then `git checkout build`?

IIRC, some versions of `git` will try and do a merge automatically when the remote branch doesn't match the local branch, which can cause this sort of annoyance when that's not what you want. I personally have set `git config --global pull.ff only`, which configures git to bail out instead of trying to do anything clever. You can run `git fetch origin` and `git reset --hard origin/build` to just reset to the remote (Github's) version instead of trying to preserve any local changes.

> If you want to adapt the _Miss Gosling_ tests to use regtest.py, I'm not at all opposed, I just don't really know how to do it. But I can take a look next week if desired.

I think these are all good ideas, though I don't think we need to figure that stuff out on this PR.

> As far as licensing, I haven't actually given much thought to it before. My immediate instinct is Creative Commons Attribution, which is what I release most of my creative stuff under, but I don't know how well that works for software. Whatever license I end up using, though, I'm happy to dual-license or whatever to ensure it's compatible with the repository license as a whole.

FWIW I don't think it's a big deal if the licence for these games conflicts with the license of dialog itself / the standard library. End users need to be allowed to to redistribute them and to run the tests, but they don't necessarily need permission to rewrite them or integrate them into their games in any way.

However, the option that involves the least thought is if you're willing to licence your work under two-clause BSD, like the rest of the repo.

bkirwi:
Your changes through 7979b1b look good to me. If you can approve the PR I'll merge this in?

dstelzer (approved):
All looks good to me! The Github integration is still a mystery to me, but this all seems to be working as intended.

bkirwi:
> The Github integration is still a mystery to me, but this all seems to be working as intended.

In case it's useful for you or some other contributor, the extremely brief summary is

`.github/workflows/build.yml` is an "action". This particular one is configured to run some commands on every PR or merge to `main`. As of this branch being merged, we have a few steps: it installs some packages, builds for linux and windows, then runs through the tests we've added here. If those tests fail it will send you an email or put a big warning on your PR... IIUC it is still possible to merge bad code anyways, but at least it'll be more likely to flag if something's broken.

You can make actions quite fancy, but for now I'd like to keep it as simple as possible and do the real work in the `Makefile`... that way it's easy to run the same tests locally, and hopefully nobody else has to think about this `yml` file very often.

PR #29, opened 2025-01-24
Fix crash when using floating divs

dstelzer:
When working on the style code, I messed up the count of local variables in some of the Z-machine runtime routines, which meant that using floating divs could crash the compiler. Oops! This fixes that.

If someone has a good testcase for floating divs, I'd appreciate that greatly; I don't fully understand how they're supposed to work, which makes it hard for me to be sure if I broke them or not.


bkirwi (approved):
I was able to repro the `dialogc` error before this patch, and it goes away afterward!

I'm going to merge quickly to get the build unbroken, so I can base a little other work on it. I'll come back to this later to see if I can understand the issue in more depth...

dstelzer:
Thanks! The code is nothing brilliant—when I updated the style code to apply to status bars too, I adjusted the way the parameters were passed to R_BEGIN_BOX_LEFT and R_BEGIN_BOX_RIGHT (the routines for floating divs), but I forgot to update one line. The result was invalid code that would crash the assembler.

Why didn't this always crash? Because Dialog has dead code elimination, so those routines were never assembled if they weren't called!

PR #30, opened 2025-01-26
Fix nonlinear unification bugs

dstelzer:
This fixes a small bug in the compiler that occurred when the same variable appears in two adjacent places in a list literal. In other words, `[$X 0 $X]` doesn't trigger the bug, but `[0 $X $X]` does. When this happened, the _latter_ instance was compiled first, but the _former_ instance allocated the memory for the variable. The result was undefined behavior: on Z-machine it would read uninitialized memory and give unpredictable values, in the debugger it would fail an assertion and crash.

Now, the _latter_ instance allocates the memory, which makes everything better.

Test case 1:
```
(program entry point)
	([$X $X] = [0 1])
	Value of X: $X.
```
Previous output:
> Value of X: 0.

New output:
[Nothing, unification fails.]

Test case 2:
```
(program entry point)
	([$X $X] = $Y)
	Value of Y: $Y.
```
Previous output:
> [$ #room]

(The second value is unpredictable and depends on what's sitting around in memory beforehand.)
New output:
> [$ $]

### DIFF
--- a/src/compile.c
+++ b/src/compile.c
@@ -627,7 +627,8 @@ static void comp_value_into(struct clause *cl, struct astnode *an, value_t dest,
 			ci->oper[0] = dest;
 		}
 	} else if(an->kind == AN_PAIR) {
-		for(i = 0; i < 2; i++) {
+		for(uint8_t j = 2; j > 0; j--) {
+			i = j - 1;
 			if(an->children[i]->kind == AN_VARIABLE) {
 				if(an->children[i]->word->name[0]) {
 					vnum = findvar(cl, an->children[i]->word);
--- a/src/compile.c
+++ b/src/compile.c
@@ -627,8 +627,7 @@ static void comp_value_into(struct clause *cl, struct astnode *an, value_t dest,
 			ci->oper[0] = dest;
 		}
 	} else if(an->kind == AN_PAIR) {
-		for(uint8_t j = 2; j > 0; j--) {
-			i = j - 1;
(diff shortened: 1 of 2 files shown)


bkirwi:
Excited for this!

It looks like this unintentionally includes a bunch of status bar related commits... okay if I rebase them out?

dstelzer:
Argh, my inexperience with Git shows again! Those are supposed to be in the `statusbar` branch instead. If you have an easy way to fix that, please do.

bkirwi:
Done! Note that if you didn't apply the suggestion [here](https://github.com/Dialog-IF/dialog/pull/27#issuecomment-2614124698) you'll want to be careful when you `git pull` next, or they'll get merged right back in.

dstelzer:
Thanks! I've done that, but I'm not entirely confident in my git branching skills still, so I appreciate you correcting the mistakes when they inevitably come back.

bkirwi (approved):
Looks good to me!

I've been trying to figure out what a good set of tests would look like for this. (It's not a natural fit for regtest.) Anyways I don't think we have to block on that to merge...

dstelzer:
Yeah, I think some testing framework akin to the _Miss Gosling_ ones would be good in general: some scaffold that compiles some code, runs it, diffs the output against the gold standard. I don't know if a general framework like that exists (at least, not one that's lightweight enough for our purposes), but it wouldn't be hard to make one, either.

PR #38, opened 2025-01-28
add style-name line to style classes in aastory output

dstelzer:
This branch adds a `style-name: whatever` line to each CSS class in the exported Å-machine file, allowing the interpreter to recover the original name of the class. A future version of the web interpreter (which I'll be handling later) can then use this to give its CSS classes meaningful names, instead of just numbers.


dstelzer:
And for once, I didn't pull any extraneous files into my commit! Huzzah!

PR #39, opened 2025-01-29
Allow JPEG cover art as well as PNG

dstelzer:
This can result in much smaller blorbs when the cover art involves photographic detail (like _Miss Gosling_ does).

### DIFF
--- a/src/blorb.c
+++ b/src/blorb.c
@@ -244,7 +244,7 @@ uint8_t *loadfile(char *fname, uint32_t *fsize) {
 int is_png(uint8_t *buf, uint32_t p_size, int *p_width, int *p_height) { // On success, loads values into p_width and p_height, and returns 0; on failure, returns 1
 	uint8_t magic[8] = {137, 'P', 'N', 'G', 13, 10, 26, 10};
 	
-	if(memcmp(buf, magic, 8)) { // Not a PNG file
+	if(memcmp(buf, magic, 8) || p_size < 18) { // Not a valid PNG file
 		return 1;
 	}
 	
@@ -274,8 +274,8 @@ int is_jpeg(uint8_t *buf, uint32_t p_size, int *p_width, int *p_height) { // Sam
 	for(;;) {
 		block = (buf[pos] << 8) | buf[pos+1]; // 16-bit block length
 		pos += block; // Skip this block (the first one will never have the image size in it)
-		if(pos >= p_size || buf[pos] != 0xff) {
-			return 2; // Malformed file
+		if(pos+9 >= p_size || buf[pos] != 0xff) {
+			return 2; // File was recognized as JPEG, but is malformed (no start-of-frame block)
 		}
 		if(buf[pos+1] == 0xc0 || buf[pos+1] == 0xc2) { // "Start of frame" block contains the image size: [ffc0] [16-bit block length] [8-bit precision] [16-bit x] [16-bit y]
 			break;
--- a/src/blorb.c
+++ b/src/blorb.c
@@ -13,6 +13,10 @@
 #include "blorb.h"
 #include "report.h"
 
+#define IMAGE_NONE 0
+#define IMAGE_PNG 1
+#define IMAGE_JPEG 2
+
 static char *strbuf;
 static int strsize, strpos;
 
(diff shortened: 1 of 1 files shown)


src/blorb.c
```diff
@@ -214,20 +223,31 @@ uint8_t *loadpng(char *fname, uint32_t *p_size, int *p_width, int *p_height) {
...
+		report(LVL_ERR, 0, "Image file %s is too small and likely corrupt: %d bytes", fname, size);
 		exit(1);
 	}
-
-	*p_size = size;
+	
+	*fsize = size;
+	return buf;
+}
+// The following files then handle different formats it could be in
```
bkirwi, line 243:
```suggestion
// The following functions then handle different formats it could be in
```

Unless I'm misunderstanding what you're getting at here!

src/blorb.c
```diff
@@ -214,20 +223,31 @@ uint8_t *loadpng(char *fname, uint32_t *p_size, int *p_width, int *p_height) {
...
+int is_png(uint8_t *buf, uint32_t p_size, int *p_width, int *p_height) { // On success, loads values into p_width and p_height, and returns 0; on failure, returns 1
+	uint8_t magic[8] = {137, 'P', 'N', 'G', 13, 10, 26, 10};
+	
+	if(memcmp(buf, magic, 8) || p_size < 18) { // Not a valid PNG file
```
bkirwi, line 247: Looks like this changed to 18 instead of 8... seems harmless but I'm curious!

src/blorb.c
```diff
@@ -238,8 +258,34 @@ uint8_t *loadpng(char *fname, uint32_t *p_size, int *p_width, int *p_height) {
...
-
-	return buf;
+	
+	return 0;
+}
+int is_jpeg(uint8_t *buf, uint32_t p_size, int *p_width, int *p_height) { // Same interface as above - see https://web.archive.org/web/20131016210645/http://www.64lines.com/jpeg-width-height for explanations
+	uint8_t magic1[4] = {0xff, 0xd8, 0xff, 0xe0}; // JPEG doesn't actually have a magic number, but the vast, _vast_ majority of JPEG files start with these four bytes, then two other bytes, then the next four
+	uint8_t magic2[4] = {'J', 'F', 'I', 'F'};
+	
+	if(memcmp(buf, magic1, 4) || memcmp(buf+6, magic2, 4)) { // Not a JPEG file
```
bkirwi, line 268: IIUC if this function was called with a sufficiently small `buf` it would overrun the buffer. I understand that's not the case - the size is checked above - but it does feel like a bit of a footgun and make me nervous about buffer overruns in general. An additional check here and whenever we bump `pos` below would make me more confident that we bail cleanly on malformed data.

src/blorb.c
```diff
@@ -256,14 +302,27 @@ void emit_blorb(
 	uint32_t imgsize;
 	int imgwidth, imgheight;
 	uint8_t *imgdata = 0;
+	int imgformat = IMAGE_NONE;
+	int status;
 
 	if(coverfname) {
-		imgdata = loadpng(coverfname, &imgsize, &imgwidth, &imgheight);
+		imgdata = loadfile(coverfname, &imgsize);
+		if(!(status = is_png(imgdata, imgsize, &imgwidth, &imgheight))) {
```
bkirwi, line 310: Overwriting the status here doesn't do anything, right? Since it's either ignored (if parsing succeeded) or immediately overwritten (otherwise)?

bkirwi (commented):
Nice change! And happy to see that blorb only supports the two formats...

src/blorb.c
dstelzer: Indeed, it's only in there for symmetry.

src/blorb.c
dstelzer: Good catch! That's what the check for 18 is for in the PNG code—to catch that sort of buffer overrun—but I forgot to do the same in the JPEG code. No harm in checking twice here as well.

src/blorb.c
dstelzer: Oop, that should be 0x18! But immediately after that we access buf[0x17], so the check is to prevent a buffer overrun if we're given a pathological file with a valid header but no data after it.

bkirwi (approved):
Thanks!

PR #40, opened 2025-01-29
Standard library support for gender-neutral language

averyhiebert:
I've implemented standard library support for gender-neutral language, parallel to the existing `(male/female $)` functionality.

### New predicates introduced:
```
(gender-neutral $)
(it $ has)
(it $ does)
(it $ doesn't)
(it $ s)
(it $ es)
```
Note that `(it $ s)` and `(it $ es)` do *not* print a pronoun, just the verb ending (analogous to `(s $)/(es $)` but intended to be used in cases where the subject is a pronoun).  This only makes a difference when the subject is singular "they".  To clarify, the intended usage is `(It $) seem(it $ s) to be harmless.`

Aside from the specific `(it $ something)` predicates, all verb conjugations remain singular for gender-neutral subjects.

### Other miscellaneous notes:

I had to introduce a *slightly* awkward special case for the "It's fine where it is" message (to avoid either "they're fine where they're" or "they're fine where they is", without introducing another new & confusingly named predicate); now if you try to GET a person who is a) using gender-neutral pronouns, and b) specified as "fine where they are," the message will be "They're fine where they're located currently."  But I don't imagine this'll come up often.

The pronoun "them" is now also updated when noticing a person with a known binary gender, since in many cases an NPC's gender may not be narrated immediately.  In other words, the following scenario is now supported:
```
You see a mechanic working on a car in the corner.
> TALK TO THEM
She ignores you.
```

Overall, the changes are relatively simple. Should be backwards-compatible with existing code.  Passes the regression tests, and I did some ad hoc testing of my own.


dstelzer:
Looking good!

While we're updating the "notice" code, this might be a good place to add an "ambiguously plural" predicate, which indicates that something should update the "them" pronoun as well as their default one. The main use for this is for inanimate things like "a pair of socks", but it could also be set by default on male and female objects.

The functionality is mostly the same as what you currently have, but it would let the author override it in ambiguous situations. If you have one character who's always referred to as female, and another who's nonbinary, it'll be annoying if THEM keeps getting assigned to the former.

stdlib.dg
```diff
@@ -4209,6 +4215,16 @@
 (its (male $))		his
 (it (male $) is)	he's
 
+(it (gender-neutral $))		they
+(them (gender-neutral $))	them
+(itself (gender-neutral $))	themselves
```
bkirwi, line 4220: I tend to hear "themself" for singular they - eg. "Shelby mutters quietly to themself." - but I understand both are in use and I'm happy with this if y'all prefer it!

bkirwi (commented):
Excited for this!

One thing we'll want to work up is a paragraph or two for the docs... I think it will be useful to help clarify the new predicates and when to use them vs. the old. If you get to writing one before https://github.com/Dialog-IF/dialog/pull/18 merges, feel free to add it as a comment here with a mention of where in the existing docs you'd expect it to slot in, and I'll make a note to merge it in to the rest of it once the dust settles.

dstelzer:
Good catch! I'd also prefer "themself", but it's a matter of style, sort of like whether or not to print a comma before the last item in a list. Probably best to leave that up to the author (by letting them edit the library file).

stdlib.dg
averyhiebert: I went with "themselves" as "themself" is still sometimes considered nonstandard, but I agree that "themself" personally feels more natural, and would support changing it if others agree.

averyhiebert:
I went with "themselves" as "themself" is still sometimes considered nonstandard, but I agree that "themself" personally feels more natural, and would support changing it if others agree.

averyhiebert:
Regarding the addition of an "ambiguous they" predicate, I agree that would be a good addition, but it might be best to address it separately just to avoid lumping too many things into one change.  I can do that and make a separate pull request when I have time later.  (I can also deal with the documentation later, as a separate pull request once the documentation merges; seems clearer than trying to convey what changes should go where in a comment)

bkirwi (approved):
Great, thanks for the followup!

PR #47, opened 2025-02-14
Fix issue with punctuation characters within lists

### LINKED ISSUE #46: [hello , world] prints as [hello, world] on Z-machine
[Details](https://intfiction.org/t/bug-with-stop-characters-in-dialog-z-machine-compiler/73665?u=draconis)

Looks like a bug in R_PRINT_VALUE where the $01 flag (which is supposed to prevent this behavior) isn't getting set properly while printing a list. That flag should always be turned on inside a list.

dstelzer:
A one-line change to the Z-machine runtime to fix #46.

The only change is in the routine R_PRINT_VALUE, which is used when printing a list. It now sets the $01 flag when recursing on the elements of that list, ensuring that they'll be printed as debugging words (with `+` marking unknown words and spaces on both sides) rather than printing words (with special spacing rules for punctuation).

Since lists are unlikely to be printed except for testing, this shouldn't lead to any changes in published games, and brings the behavior in line with the debugger and the documentation.

### DIFF
--- a/src/runtime_z.c
+++ b/src/runtime_z.c
@@ -530,7 +530,9 @@ struct rtroutine rtroutines[] = {
 			{Z_PRINTLIT, {}, 0, 0, "["},
 			{Z_STORE, {SMALL(REG_UPPER), SMALL(0)}},
 			{Z_STORE, {SMALL(REG_SPACE), SMALL(1)}},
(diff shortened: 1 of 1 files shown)


bkirwi (approved):
Looks good!

Feel free to delete the old version of the line, if you like... we can always revert this change if we need to.

PR #49, opened 2025-02-14
Make Unicode characters in multi-character dict words a warning instead of an error

dstelzer:
The Z-machine backend doesn't allow non-ZSCII characters in dictionary words, and previously that was a fatal error. But sometimes the compiler was overzealous about deciding which words belonged in the dictionary, and would disallow perfectly legitimate programs, just because it couldn't prove those words _wouldn't_ somehow end up in the parser.

In particular, if _any_ closure is used in parsing, the compiler can't guarantee that _all_ closures won't be used in parsing. As a result, a Unicode character in any other closure—even one nowhere near the parser—would be forbidden. For example:

```
(program entry point)
	*(repeat forever)
	(line)
	(get input $Words)
	(collect $Obj)
		(determine object $Obj)
			*(object $Obj)
		(from words)
			*(dict $Obj)
		(matching all of $Words)
	(into $List)
	$List
	(query {ašdf})
	(empty $Words) %% Fail and repeat if not empty

#apple
(dict *) apple red fruit

#pear
(dict *) (query {pear})
```

This PR changes that error to a warning, and generates a dictionary word with the Unicode characters replaced with periods. This is a trick used in Inform 6 to create words that can never be parsed, without messing up the Z-machine's dictionary. Now, the compiler will instead warn:

> Warning: Unsupported character U+0161 in dictionary word `@ašdf`. This word will never be recognized by the parser.

But if the author is confident that `@ašdf` will never need to be parsed—like in this case—compilation can continue without a problem.

### DIFF
--- a/src/backend_z.c
+++ b/src/backend_z.c
@@ -278,16 +278,17 @@ static uint8_t unicode_to_zscii(uint16_t uchar) {
 	}
 }
 
-static int utf8_to_zscii(uint8_t *dest, int ndest, char *src, uint32_t *special) {
+static int utf8_to_zscii(uint8_t *dest, int ndest, char *src, uint32_t *special, int unicode_halt) {
 	uint8_t ch;
 	uint32_t uchar;
 	int outpos = 0, inpos = 0, i;
+	if(special) *special = 0;
 
 	/* Stops on end of input, special unicode char, or full output. */
 	/* The output is always null-terminated. */
 	/* Returns number of utf8 bytes consumed. */
 	
-	/* Experiment: don't stop on special unicode char, make it a warning instead of an error, because the compiler can sometimes be overzealous with these errors and disallow perfectly valid programs if it can't prove certain words are *not* used for parsing */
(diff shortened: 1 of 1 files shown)


dstelzer:
This fix only works for multi-character words at present. I'll need to figure out a solution for single-character words.

dstelzer:
Indeed!

So, by default, the `utf8_to_zscii` function, if it encounters an illegal Unicode character, sets `*special` to the codepoint of that character and returns. This is because `utf8_to_zscii` is used both for object names and dict words (if `*special` is set, crash) and for strings (if `*special` is set, compile a PRINT_UNICODE opcode, then continue after that character).

I added an extra parameter to the function that, if set, will make it _not_ stop on an illegal Unicode character; it'll put a dot in that position and move on. This meant I needed to make it set `*special` to zero at the beginning of the function, rather than right before returning successfully—since I do still want to return that information!

That flag is set only when handling dictionary words, so the behavior won't be changed anywhere else. Producing dictionary words with dots in the middle of them will ensure they can't be parsed (an old I6 trick) but without changing the structure of the dictionary.

bkirwi (approved):
Aight! Still more comfortable if we had test coverage of this, but it seems fine if we want to handle that in a separate PR.

dstelzer:
Yeah, I have a test case for it, but I'm not quite sure how to integrate it when the tests are all in a different branch. But I'll add it to the more-tests branch for now!

PR #50, opened 2025-02-17
Add a doctest script

bkirwi:
This adds a script that performs very basic validations of the code in the docs. In particular, it asserts that code blocks can be loaded by `dgdebug` without errors... or, if it's an example that's expected to error, it should complain if `dgdebug` loads the code successfully.

I've also made some changes to the docs themselves:
- Sometimes the code was invalid; those are now fixed.
- Sometimes the code is only valid in the body, but not at the top level. Sometimes we indent those code examples; sometimes we don't. We now always indent such code, and the test script ignores code that starts with an indent.
- Some examples are examples of things not to do. (Infinite loops; invalid syntax; re-declaring builtins.) These are now explicitly annotated. (In a way that hopefully shouldn't change the generated website, though I don't currently have a way to check.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,5 +7,6 @@ test/regtest/%.zblorb: src/dialogc test/regtest/%.dg stdlib.dg
 test: test/regtest/cloak.zblorb src/dgdebug
 	bin/regtest.py -v --game test/regtest/cloak.zblorb --interpreter dfrotz test/regtest/cloak.regtest
 	make --directory=./test/gosling test
+	bin/test.py doc
 
 .PHONY:		test src/%
--- a/bin/test.py
+++ b/bin/test.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 import os
 import argparse
+import re
 import subprocess
 import tempfile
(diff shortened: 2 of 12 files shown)


bkirwi:
I originally intended to write tests that verified the `output` blocks, but:
- The simpler version already found a few issues.
- That's a bit redundant with some of the work @dstelzer has been doing adding fine-grained tests based on the docs, and I didn't want to preempt it. (Though even with the maximal version of that, checking the docs might still have some value, since it verifies that the example code we provide does what we expect.)

I'd still consider doing that in a followup. I'm also considering improving coverage by taking the body-fragment blocks we currently skip, wrapping them up in a rule, and then passing that whole rule to `dgdebug`. Undecided!

bkirwi:
Ugh, never mind... got slightly stuck on this and ended up adding output assertions too. Right now these are mostly useful for `lang`, since `lib` tends to use `output` for little snippets of interaction and I don't have support for that yet.

dstelzer (approved):
This all looks good to me! The documentation changes are minor and improvements, so I see no reason not to merge.

PR #51, opened 2025-02-20
Contributor's guide and version update

bkirwi:
Writing down some general project guidelines for contributors.

Versioning in particular has been discussed in #16, https://intfiction.org/t/dialog-version-naming/73515 and elsewhere. This is my current best attempt at summarizing the rough consensus. I can follow up later with more details as we settle on them, including the release process.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -1,0 +1,51 @@
+# Project guidelines
+
+## Languages and tools
+
+The core Dialog tools (`dgdebug` and `dialogc`) are implemented in C and built
+with GCC and Make. We'd like to keep this source as portable as possible.
+
+Additional tooling, such as for testing, is implemented in dependency-free
+Python 3. Users shouldn't need to run Python code to build the core tools, but
+contributors may need them to run tests etc.
+
+The documentation is in Asciidoc. The public-facing docs are currently built
+using Antora, and that build process is not part of this repository, and the
+docs themselves do not use Antora-specific syntax or extensions.
+
+## Versioning
+
+There are a few different versions at play here:
+
+- **Language** - The language version is a two-character string, like `1a`. This
+  version changes when the Dialog language itself changes. (As a rule of thumb,
+  it's a language change when existing code might break, or when we need to
+  orchestrate changes across both the compiler and library.)
+- **Compiler** - The compiler version is a combination of the language version
+  and a two-digit suffix, like `1a/00`. This version applies to both `dialogc`
+  and `dgdebug`, and increments whenever a new version of these tools is
+  released.
+- **Library** - An old-fashioned version string like `1.0.0`. The first two
+  parts correspond to the two digits of the language version, so the first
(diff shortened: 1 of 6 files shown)


dstelzer:
Looks good to me overall! Just a few things:
- Previously, the patch numbers for the compiler have started at 01, so this would be 1a01 instead of 1a00; or do we want to change that going forward? (No strong preference)
- The library minor versions have also skipped past l and o, to avoid them looking like 1 and 0; do we want to continue that practice? (No strong preference) If so, do we also want to skip minor versions 12 and 15 of the library? (I vote yes)
- In the `more-tests` branch I have a step in the Makefile that removes the compiler and library version from the output, so that test scripts won't break every update; should we do that to the Gosling and Impossible test cases? (I vote yes)

bkirwi:
> Previously, the patch numbers for the compiler have started at 01, so this would be 1a01 instead of 1a00; or do we want to change that going forward? (No strong preference)

Hmm, I don't have a strong argument either way, so I guess I should probably go with "what Linus did". I'll switch it unless I think of something more compelling than "one extra number" or "slightly more normal for software".

> The library minor versions have also skipped past l and o, to avoid them looking like 1 and 0; do we want to continue that practice? (No strong preference) If so, do we also want to skip minor versions 12 and 15 of the library? (I vote yes)

This one bugs me on an aesthetic level, but I also don't have a great argument for changing tack. But at least this one we can punt until ten breaking releases or so...

> In the `more-tests` branch I have a step in the Makefile that removes the compiler and library version from the output, so that test scripts won't break every update; should we do that to the Gosling and Impossible test cases? (I vote yes)

Normally I prefer to change to tools to make regenerating these files very easy (ie. one command), since this stuff doesn't change _that_ often and removing it makes the test coverage slightly worse. But I can't say I feel strongly in this specific case! Happy to accept such a change if you want to submit it.

dstelzer:
That is true, it would make a small hole in the test coverage, but we can add another test that specifically ensures `(compiler version)`, `(library version)`, and `(serial number)` all work properly. I think needing to update the test cases every time we bump the patch number will be more annoyance than it's worth. (The Makefiles for `/test/simple/` specifically dummy out all those predicates so they'll always print DEBUG, to avoid the hassle.)

I don't think avoiding `l` and `o` is really that big a concern, either way, but you're also right that we won't have to worry about it for quite a while.

I'd also be inclined to have the minor versions start at 1, because "a = 1" is more obvious to most people than "a = 0".

bkirwi:
> I think needing to update the test cases every time we bump the patch number will be more annoyance than it's worth.

Cool! Again, don't feel that strongly - happy to go with whatever you have for now.

> I'd also be inclined to have the minor versions start at 1 [...]

Sure, changed. I've also updated the initial version to `1a/01`, like you suggested up top.

I think that probably makes this okay to merge?

dstelzer (approved):
Looks good to me!

PR #53, opened 2025-02-22
Build for MacOS and generate a release ZIP

bkirwi:
- Builds and tests on MacOS as well.
- Generates a ZIP which more-or-less approximates Linus' release zips. Differences include a change in supported architectures (intentional) and including the documentation source instead of the generated HTML (unfortunate, but I'd rather not block this until the good stuff is available).

This theoretically will only generate a release if a `release-XXX` tag is pushed. So for example `release-1a01-1.0.0` would generate `dialog-1a01-1.0.0.zip`.

However we do upload the zip for every build, if you want to inspect it. It's a little awkward to navigate to, but if you click the `build-release` link in the checks below, and then the ":house: Summary" link at the top left, the "archives" generated by the build process are downloadable at the bottom. (Though for annoying technical reasons the release zip is itself wrapped up in another zip file... oh well.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,15 +2,15 @@ name: Build
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ['main']
+    tags: ['release-*-*']
   pull_request:
-    branches: [ "main" ]
+    branches: ['main']
 
 jobs:
-  build:
 
+  build-linux:
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v4
     - name: install dependencies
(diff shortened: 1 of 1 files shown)


bkirwi:
Oh, and I should say... hard to test the actual release process before it's merged, so expect a follow-up with fixes later. But I think the basics at least should be in good shape...

dstelzer (approved):

bkirwi:
Thank you for taking a look!

> I don't really know enough about Git build tools to understand the mechanics [...]

On the off chance you're interested - this workflow stuff is GitHub-specific, and the single most helpful thing I've found (while working on this) is this cheat sheet: https://github.github.io/actions-cheat-sheet/actions-cheat-sheet.html

Overall I find it a little fussy... but it does mean that now every PR will run the full set of tests on Linux and Mac, then check that it can build the full release zip, and that would be a big pain to arrange otherwise I think, so I can't complain too much!

PR #56, opened 2025-03-06
Add more regression tests covering the language, the library, and our new features

dstelzer:
This isn't really finished yet, but I promised I would push what I had on Wednesday, and Wednesday has come.

This contains:
- A test suite in `test/simple/language` that covers various aspects of the language. `make all` in this directory will run all these tests on both Z-machine and debugger. It can be extended to cover the Å-machine too if we can get the node.js Å-machine interpreter working.
- A similar test suite in `test/simple/library` covering the library. This one needs significantly more work.
- A test suite in `test/remglk` that's meant to test fancy output that doesn't come across in dumbfrotz. This one doesn't have a Makefile yet.

Right now, all the tests pass using Linus's compiler. I haven't yet updated them for some of the fixes in this repo, which will affect the output in some cases. And right now I'm assuming the version of frotz that's in apt, which has various bugs—not showing the status line, for example, and not including any text that's styled differently from default (which is in fact the cause of not showing the status line).

Still, I think this is a good stage for other people to look at it and give feedback. And if you'd like to get the library tests working properly, or find a way to run the node.js Å-machine interpreter or emglken for the RemGlk output, it would be very much appreciated. I've somewhat hit a wall with working on this particular aspect of the repo, and can't seem to make any more headway on it.


bkirwi:
Wow! That's a lot of tests!

I propose we just merge this pretty much as-is, and treat any feedback as followups. I'd already expect it to provide a lot more coverage than we have now. I do have some vague ideas around how maintaining these could be made less of a hassle, but I think it may make more sense to just send those as followup-issue PRs instead. (If they happen to work out, which is not guaranteed.)

With one exception - can any of these be added to the top-level `make test`? That way any PRs we merge in the meantime will have slightly more coverage...

And in any case thanks for your work on this! Big job.

dstelzer:
They can definitely be added to `make test`, and I think they should be once they all pass! Right now there are only four that are incomplete—8b, 9c, and 11b need their input specified, and 10a needs to be written—and those can be dummied out for the moment.

We just need to decide which version of dfrotz we should be targetting, since that affects all the output.

dstelzer:
I've now added these tests to the main Makefile. Some of them are unfinished, but they all pass with the version of the source currently on this branch.

Note that some of them will need to be updated once we merge, because they're testing behavior that's glitchy in the version of `main` that this branch was forked from, but is fixed in other PRs. But that can be done as part of those PRs.

dstelzer:
Ah, something seems to have gone wrong with the version number. Anyone know what the issue is here? I tried to avoid interfering with that in this PR…

That aside, the three big things I'd like to add, test-wise:
- Use the node.js Å-machine interpreter to run all these tests on Å-machine too
- Use the node.js emglken interpreter to run a couple of style tests on RemGlk
- Fill out the three remaining TODO tests (fungible objects, choice mode, and dialogue)

But I'd like to get this one merged so that we can do our first release, and then I (or even better, someone else!) can hit those later.

dstelzer:
I've fleshed out the remaining tests; that should cover the whole language and library! I'd still like to add some RemGlk tests for styling, but I think this is solid enough to merge now.

…the problem is, my attempts to hook this into the main Makefile test script seem to have backfired. I can't figure out why tests are failing now that I didn't touch at all, like the version number in the Impossible Stairs test—I didn't update the version number, it should still be 1a/01-dev.

I'd appreciate some more eyes on this, because I can't figure out why it's suddenly breaking.

dstelzer:
Of course, I immediately realize what's going on after asking this. Updating the Makefile meant the Impossible Stairs test got incorporated for the first time, and it fails due to a few changes that have been made to the library.

In particular, when you walk into a room with a person in it, "them" now gets updated to point to the person (due to #40), which means it's _not_ pointing to the object it previously did. Do we want to update the test, or adjust the library?

dstelzer:
Aha, I think I found why the build is failing specifically on Mac—the Mac version of the `sed` command has different syntax from the POSIX one!

This is going to be a headache… @bkirwi would it be possible to install gnu-sed via Homebrew and `alias sed=gsed` before running the tests? That would give a sed variant that works like the POSIX one, letting the test scripts work properly.

Otherwise, I'll need to find some sed alternative that works on both OS X and POSIX. A Perl script or something.

dstelzer:
It looks like replacing sed with perl shouldn't be too hard, in theory, so I'm going to give that a shot. We'll see if it works!

dstelzer:
Ugh, replacing sed with perl worked, but now I'm running into issues where OS X diff doesn't support the same options as GNU diff.

@heasm66 , you have a Mac system, right? Could you take a look at why the diff commands in test/simple/language/Makefile don't work on OS X? The best I can do is try something, push, and see where the build fails, which isn't really an ideal way to test it.

bkirwi:
I've pushed a small makefile fix to this branch, and tried installing a newer `diff`.

It looks like that worked, so I'm going to be slightly brazen and go ahead and merge. It's already a huge improvement to have all these tests on every build, and I think we can take any further improvements as follow-up PRs.

bkirwi (approved):

PR #57, opened 2025-03-06
Update changelog for new release

dstelzer:
I'll also use this branch to update the documentation before release, but first we'll have to figure out how to get the documentation into this repository in the first place!

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -34,9 +34,60 @@ mingw32):
 
 Project website:
 
-	https://linusakesson.net/dialog/
+	https://github.com/Dialog-IF/dialog/
 
 Release notes:
+
+	1a/01, Lib 1.1.1:
+	
+		This is the first release handled by the community, and per
+		Linus's wishes, we're bumping the major version for the first
+		time! We're also taking this opportunity to unify the version
+		numbers between the compiler, library, and manual.
+		
+		Backend: color and background-color properties are now supported
+		on Z-machine backend.
+		
+		Backend: styles can be stacked on Z-machine backend, allowing
+		for (for example) bold italics, or monospace reverse-video.
+		Spans are also allowed inside status bars as a result.
+		
+		Backend: fixed a bug that would crash the Z-machine if too many
+		dictionary words were included in a wordmap, pushing the
+		wordmap datatables past address $8000.
+		
+		Backend: lists containing punctuation marks will now print with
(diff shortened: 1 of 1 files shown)


dstelzer:
I've started working on the documentation as well, but I would especially appreciate additional eyes on that part. I believe I've documented all the changes to the language's syntax, but I don't know how clear the style is.

manual/modules/lang/pages/io.adoc
```diff
@@ -320,6 +334,29 @@ Note that if more output is sent to the cleared div, this new output may or may
 All of the above predicates succeed (except when they generate a runtime error).
 All of them may be ignored by interpreters.
 
+[#dimensions]
```
bkirwi, line 337: Possible I missed it, but... is this actually implemented? (If not, I'd love if we could make this change as part of that future PR... that way we don't have to remove it temporarily if it slips a release.)

bkirwi (commented):
One question, but overall this is great!

manual/modules/lang/pages/io.adoc
dstelzer: The (current div width $) and (current div height $) ones? Huh, I could have sworn I made a PR for those, but it seems like I never did!

My bad; that was implemented back in January, but I never submitted it at the time. Now it's in #77 and we'll see if that causes any issues. If it does, I'll switch these to that branch.

bkirwi: Alright! I'll definitely want to give that change some thought, but I'd be ready to merge this PR otherwise... up to you.

dstelzer: I'd prefer to merge this along with #77 , once any issues there have been dealt with; it's the sort of change I'd like to do when bumping the minor version number, and it feels odd to do that immediately after a new major version.

bkirwi: Got it. To be up-front: I have some qualms about the design and semantics of #77 and similar (mostly drawbacks you've already identified) but don't feel like I have the background to really discuss it in a meaningful way, and I don't have a great sense of when that might change. I definitely don't want to block progress in that area, but I'm also unlikely to be much help with it anytime soon!

dstelzer: That makes sense! I'll open a new issue to discuss it, rather than here.

PR #60, opened 2025-03-06
Fix error with serial commas

### LINKED ISSUE #55: Support the serial comma in commands
Inform famously chokes when given a command like TAKE APPLE, BANANA, AND CANTALOUPE. Unfortunately, so does the Dialog standard library at present. But while the Inform issue is likely to never be fixed (the parser is simply too old for changes like that to be viable), it should be solvable in Dialog.

dstelzer:
Previously, using the serial comma led to a parse failure:

> \> take apple, pear, and banana
> You take the apple.
> 
> (I'm sorry, I didn't understand what you wanted to do.)

This is because when parsing a noun list, it splits on the first instance of `@,` or `@and`, then tries to parse the left side as a single object and the right side as potentially multiple objects. With a serial comma, this leads to an empty left side, and thus parsing fails.

A one-line change to the standard library (well, technically two line change) strips a leading `@and` off before parsing, which fixes this issue.

> \> take apple, pear, and banana
> The apple: You take the apple.
> 
> The pear: You take the pear.
> 
> The banana: You take the banana.

This fixes #55 .

### DIFF
--- a/stdlib.dg
+++ b/stdlib.dg
@@ -5316,6 +5316,10 @@
 	(object $Head)
 (mark multi-object $Other into $Other)
 
+%% Avoid error when using serial comma (TAKE X, Y, AND Z)
+(parse noun list [and | $Rest] as $ObjList $Policy $AllAllowed)
+	(parse noun list $Rest as $ObjList $Policy $AllAllowed)
+
 (parse noun list $Words as $ObjList $Policy $AllAllowed)
 	(split $Words by [, and] into $Left and $Right)
 	*(parse basic noun $Left as $LeftObj $Policy $AllAllowed)


dstelzer:
@bkirwi I think this may be an issue with diff, since it's only happening on Mac, and only happening in the Gosling test, which doesn't use the standard library.

But I'm not sure why this branch would show this failure, and none of the others. This branch doesn't touch any of the files relating to Gosling, so why would that one suddenly malfunction?

bkirwi:
I rebased the change so we got the test changes etc. that just merged, and the complaint has gone away... so I suspect you're right that it's related to diff somehow. It probably has to do with this branch being based on a version of the code with an older version of the action, but I don't feel compelled to look into it that deeply unless it crops up again...

bkirwi (approved):

Issue #34, opened 2025-01-27
Handle Unicode characters in closures on Z-machine


dstelzer:
The Z-machine can output any Unicode characters in the BMP, but can't handle most of them in input. If these characters are used in a closure, though, the compiler can't guarantee that they _won't_ be used for input, and errors out just to be safe.

Since the author _can_ guarantee that a closure isn't used for input, this should be made a warning instead.

bkirwi:
What would happen if the author is wrong, and they are used as input somehow?

dstelzer:
I need to study the Z-machine specification a bit more for that. But the short answer is either:
- The interpreter strips out the illegal characters, so the resulting word is recognized, but looks wrong if printed back
- The interpreter doesn't allow the illegal word through to the game, so it'll never be recognized at all

dstelzer:
It shouldn't be able to cause a crash or anything, though—those characters just don't appear in the game's character set, so there's no way to represent them in the parse buffer, and it's the interpreter's job to fill the parse buffer with something meaningful.

dstelzer:
With the fix in #49, the result is dictionary words that can never be matched by the parser, but shouldn't cause any problems.

dstelzer:
Handled in #49

PR #64, opened 2025-03-06
Adjust library to circumvent optimization issue

### LINKED ISSUE #63: Avoid code paths that could query a closure inside a (collect words)
Dialog's optimizer struggles with closures, and if any codepath invoked within a (collect words) could query a closure, every word ever called by a closure will end up in the game's dictionary. This is the source of #34 : the `(disambiguate by object name...)` predicate calls `(the full $)` within a `(collect words)` (technically in this case a `(from words)`). `(the full $)` then queries `(the $)`, which calls `(listing...)` if it's provided with a list.

However, `(the full $)` is only ever called with individual objects in this code path. Duplicating the part of `(the full $)` that handles individual objects and calling _that_ in this instance will allow the compiler to optimize closures much more effectively.

dstelzer:
The optimizer currently chokes when any codepath leads from a `(collect words)` or `(from words)` to a closure query, whether or not that codepath is ever executed. For the most part, the library is very good at avoiding this. But there's one place where it isn't, and that means that code with lots of closures is not optimized as well. This is at the root of #34 .

With a few adjustments we can avoid presenting that never-executed codepath to the compiler, improving the optimization results significantly. The library code is a little less elegant now, but there are several places where the library specifically chooses to help the optimizer over elegant code.

This fixes #63 .


dstelzer:
A test case:

```
(current player #player)
(#player is #in #room)
(room #room)

(instead of [wait])
	(query {Ỳőŭ ŵåıṯ.})
```

Compile this with stdlib.dg, targetting the Z-machine (Z5, Z8, or Zblorb). With Linus's compiler, it'll produce an error; with our updated compiler, it'll produce a warning.

After these changes to the library, that error or warning is gone, since the optimizer can now prove that this closure is never queried inside a `(collect words)`, and thus these strange symbols don't need to be added to the dictionary.

bkirwi:
I've closed #66, which appears to be AI-generated spam that was based on this PR. (What a world we live in...) Not relevant to this PR, but I figured I should mention it somewhere public.

bkirwi:
> With a few adjustments we can avoid presenting that never-executed codepath to the compiler, improving the optimization results significantly.

Overall on board! Questions:
- Can you elaborate on or quantify "significant"? (What good impacts does a user see from this change?) I don't have a great sense of the optimizer behaviour, so I'm fuzzy on exactly what extra work we're avoiding...
- It seems like Linus made some effort to avoid this case, but introduced it anyways. Seems like it'd be nice to have some way to make sure that we don't reintroduce a similar problem in the future... can we get a test around it somehow? I suppose we could add an example like the one you shared, then assert that there's no warnings in the output... but maybe there's an easier way.

dstelzer:
Right! So, when it comes to printing actual literal words, there are two different ways Dialog can do it. If it's in a codepath that can be executed in a `(collect words)` or `(from words)` context, it adds all those words to the dictionary, then compiles a "print dictionary words" opcode. Otherwise, it just prints it as a literal string.

The latter option is more efficient in that it avoids cluttering up the game dictionary, produces smaller files (strings are compressed more than dictionary words), and also can handle Unicode in a more efficient way—for Unicode (that is, non-ASCII) characters to be included in dictionary words, they have to be added to the game's character set, which has very limited space. On the Z-machine, it'll just crash with Linus's compiler, and produce a bunch of spurious warnings once #34 is fixed.

Normally, the optimizer is very good at handling this. But if _any_ closure is ever queried inside a `(collect words)` or `(from words)` context, then _every_ closure is assumed to be queried there—the compiler can't keep track of what happens to closures after they're created, so it assumes every closure can potentially be called in every `(query $)`.

This caused problems for _Miss Gosling_, where I had an `(emph $)` predicate that takes a closure, sets the appropriate styles, then queries that closure. This meant that various Unicode punctuation marks were included in closures, which made the compiler choke. But even after downgrading that error to a warning, it leads to hundreds of extra words being put in the game dictionary when they're not needed—which is a problem when the size of the dictionary is finite, and each of those words uses up precious file size!

I'll add a test case for it; this should do it:

```
(current player #player)
(#player is #in #room)
(room #room)
(look #room) (query {¡Ħěłĺø ŵőřľđ!})
```

If any closure can be queried inside a `(collect words)`, then on the Z-machine backend, this will produce a warning or error saying that those Unicode characters are not valid in dictionary words. Otherwise, it'll compile just fine.

I think I want to tweak my code for this a bit more, though, because right now it involves duplicating some things that would be better not duplicated. I'll toy with it more today if I have the time.

bkirwi:
Ah, I didn't realize the implications for the dictionary. That makes a ton of sense, especially on the Z-machine. And your test plan sounds good!

dstelzer:
I've made a more elegant implementation now, with less repeated code. It also passes the test case, though I'm not sure how to add that test case to this commit when all the testing infrastructure is in another branch. (For now I'm just testing it by hand with `dialogc -t z5 closures.dg stdlib.dg`.)

bkirwi:
I think the testing here is unblocked now that the testing stuff is in!

Reading through again and looking through the docs... would this mean that users would need to switch to defining these new predicates, like the library does?

dstelzer:
For the most part no—even if you define a custom `(the $)` for your object, it will only cause a problem if that custom `(the $)` introduces a new word that should be recognized during disambiguation prompts, but should _not_ be recognized during normal parsing. And that seems like enough of an edge case not to worry about.

It is a fair concern, though. Do you think it should be mentioned in the documentation?

bkirwi:
Fixed a merge conflict; hope you don't mind.

It appears that tests are failing; seems like we're including somewhat less disambiguation information.

> It is a fair concern, though. Do you think it should be mentioned in the documentation?

I think it's fine? I've been toying with changing the names a bit, so `(the $X)` is single and the new predicate `(the plural $X)` replaces the existing predicate and gets defined for lists etc. But I don't know that it really improves ergonomics, and as you suggest it's unlikely to ever come up.

dstelzer:
Ah, thank you; I still struggle with a lot of features of Git, as you can see. It wasn't my intention to include less disambiguation info, so it's good that the tests are catching it!

dstelzer:
Oho, I see what's going wrong. That test case overrides `(the full $)` to add more details to show how confusing it can get.

dstelzer:
I've changed the test case to now override `(the full single $)`, but this suggests we should put a note in the documentation about it.

I also changed the test case that currently expects warnings about closures in `(from words)` to no longer expect that, since, well…it's fixed now!

dstelzer:
Oops. I thought there was a simpler solution; there is not.

I've added a note to the documentation, since now we know it comes up in one of the doc examples. I think that should be everything now!

bkirwi (approved):
Reviewed the followup! Seems like a reasonable tradeoff in the end; thanks for the docs etc.

dstelzer:
This is less of an issue now that it doesn't cause the Z-machine backend to choke completely, but it'll still help with optimization.

Issue #22, opened 2025-01-22
Separate meta text from in-game text


dstelzer:
By default, the library makes no distinction between in-game text ("the door is closed") and meta text ("that's not a verb I recognize"). But unlike Inform, Dialog has a style system that makes it straightforward to add such a distinction. If we wrap a `(div @meta)` around the library messages that come from the system instead of from the model world, the player can then tweak the styling to their preference, and take advantage of the flexibility of divs and spans.

PR #65, opened 2025-03-06
Add (div @meta) to meta text in stdlib.dg

dstelzer:
By "meta text" I mean text that comes from the program, rather than from the game world.

There's sure to be disagreement about how this should be implemented; for now I've used divs, for example, but perhaps spans would be less disruptive to the line spacing of existing stories. But I wanted to submit an initial pull request so that people can play around with it and see how it feels, and weigh in on how they'd prefer it to work.

This addresses #22

### DIFF
--- a/stdlib.dg
+++ b/stdlib.dg
@@ -3059,13 +3059,11 @@
 	(div @meta) { Really quit? \(y (no space) / (no space) n\) }
 	(if) (yesno) (then)
 		(div @meta) (display quit message)
-		(line)
 		(quit)
 	(endif)
 	(stop)
 
 (display quit message)
-	(par)
 	Thanks for playing!
 
 %% RESTART
@@ -5038,7 +5036,6 @@
 	}
 
 (game over menu)
-	(line)
 	(div @meta) {
 		Would you like to:
 		(if) (interpreter supports undo) (then)
@@ -5834,7 +5831,7 @@
 			(div @meta) {
 				\( I only understood you as far as wanting to
 				(describe action $BadAction)
-			. \)
+				. \)
 			}
 		(endif)
 	(else)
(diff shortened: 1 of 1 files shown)


dstelzer:
Hm. This currently fails all the tests because now the meta text is disappearing on dfrotz (due to the style bug). Should we be changing the library to work around a bug in a testing tool, though?

stdlib.dg
```diff
@@ -6455,11 +6488,15 @@
 %% Fatal runtime errors
 
 (error $Code entry point)
-	(roman) \( Technical trouble:
-	(report fatal error $Code)
-	Attempting to recover with UNDO. \) (line)
+	(roman)
```
bkirwi, line 6491: I think perhaps this never does anything now, since we immediately enter a div that sets font to italic?

bkirwi (commented):
Overall a good direction I think!

I do think that `div`s seem like the right call here; it seems like we're never interleaving these "meta" responses with other text in the same paragraph, and at least for the stories I have in progress this would give me more desirable options for styling.

bkirwi:
Er, yeah, and it looks like `dfrotz` is failing for different reasons on different builds... Linux due to the bug you mention, and MacOS because it's emitting status lines whereas it wasn't before.

Do we like having the status line in the output? If so I can probably try and get the linux build on a newer version. If not I can pin MacOS to an older version until it becomes optional...

stdlib.dg
dstelzer: That's correct. No reason not to remove it now!

dstelzer:
So that unfortunately gets complicated. Because you see, dfrotz was _always_ meant to output the status line—but the status line is by default set to reverse-video style, and divs using styles weren't being output! This means there's no version where styled divs appear but the status line doesn't.

I think the best thing to do would be to upgrade to a new version of dfrotz, so that we don't have to keep working around this bug forever, but it'll mean going through and updating every test to include the status line.

dstelzer:
I _think_ all the tests should be fine once #69 is merged—things are working okay on my machine, at least—but there may be some line breaks to check on, because we've inserted new divs.

dstelzer:
There, it looks like it now passes the checks properly!

bkirwi (approved):

PR #69, opened 2025-03-09
[wip] Miscellaneous build fixups

bkirwi:


bkirwi:
There were a couple more minor things I wanted to sort out, but I'm not sure I'll get to them tonight... putting up what I have.

dstelzer:
This all looks good to me! The only issue I can imagine coming up with this is if the homebrew version of dfrotz diverges from the most recent version in gitlab; is there a reason not to use homebrew for both?

But the timespan between a new version appearing on gitlab and on homebrew is not large, and if a change comes that'll affect our testing, we'll have to account for it anyway. So it's not that big of an issue.

bkirwi:
> The only issue I can imagine coming up with this is if the homebrew version of dfrotz diverges from the most recent version in gitlab [...]

Yeah, agreed. Ideally we should pin the version for both... in theory you can install specific versions via homebrew also, but I couldn't get it working quickly. I will come back to this if it ends up being an issue!

dstelzer:
If it does, it'll be an edge case of an edge case. So I see no reason not to approve!

dstelzer (approved):

Issue #23, opened 2025-01-22
Reclaim memory before the final question


dstelzer:
Code that runs during the final question (asking if you want to restart, restore, undo, or quit) can very easily exhaust the heap, because this question is run in the middle of action processing. If this ran outside of action processing instead, all that memory would be reclaimed, and more elaborate processing would be possible.

PR #71, opened 2025-03-13
Backtrack to free memory before final question

dstelzer:
Dialog's dynamic memory model is fairly simple. It has no garbage collector, so it only frees memory when backtracking (due to `(fail)` or `(stop)`—the two ways of calling the Z-machine `@throw` opcode in Dialog).

This means that doing multiple parses without backtracking in between can quickly exhaust the heap. This originally became a problem with disambiguation questions, so the standard library (several versions ago) was updated to backtrack before disambiguating and reclaim that memory for the next parse.

This was not done, however, for the final question ("Game over! Do you want to undo, restart, restore, or quit?"), because the parsing there is simplistic enough that it generally doesn't cause a heap overflow. But if any complicated code is added to the final question, it will quickly crash the interpreter—this happened with the automap in _Familiar Problems_.

Since the backtracking code was already added for disambiguation questions, it takes only a couple lines of code to make the final question use it too. Now you can get as complicated as you like during the final question, without overflowing the heap.

### DIFF
--- a/stdlib.dg
+++ b/stdlib.dg
@@ -4776,7 +4776,7 @@
 			(parse commandline $Words with choices [])
 		}
 	(elseif) (game has ended) (then)
-		(actual game over)
+		(game over)
 	(else)
 		(par)
 		(stoppable) {
--- a/stdlib.dg
+++ b/stdlib.dg
@@ -4988,7 +4988,7 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %% Game over
 
-(actual game over) %% The (game over) predicate first backtracks to free memory, then calls this
+(game over) %% The (game over $) predicate first backtracks to free memory, then calls this
 	(par)
 	(game over status bar)
 	(if) (scoring enabled) (then)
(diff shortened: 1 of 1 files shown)


dstelzer:
This fixes #23 .

stdlib.dg
```diff
@@ -5003,6 +5005,10 @@
 
 (interface (game over $<Closure))
 
+(game over)
+	(now) (game has ended)
+	(stop)
```
bkirwi, line 5010: It looks like `(game over)` is only ever called in one place currently... could we inline this there and avoid needing to add a second predicate?

bkirwi (commented):
Looks like this would be simple to do without changing the actual `(game over)` predicate itself, so I'd lean towards that if you don't mind it...

Like and appreciate the change otherwise!

dstelzer:
My original rationale was that `(game over)` was part of the public API, so I didn't want to affect its functionality too much. But now I see that it's not actually documented in the manual, only `(game over $)`, which means we can change it if we want!

stdlib.dg
dstelzer: My original rationale was that `(game over)` was part of the public API, so I didn't want to affect its functionality too much. But now I see that it's not actually documented in the manual, only `(game over $)`, which means we can change it if we want!

dstelzer: And I've now done so. No need for `(actual game over)` any more.

bkirwi (approved):
Great!

PR #73, opened 2025-03-17
CI for the manual

bkirwi:
This tackles the "how do we check the manual in CI" and "how do we generate a version for release" questions we've had by just running a default build with pretty basic settings.

This is not as pretty as the real custom site, and does not have all the same features... but it does at least do some basic error checking, and it gives us something to stick in the release zip. I've put `Dialog (snapshot)` in the header to make it more clear that this is not the Full Documentation Experience, however.


dstelzer:
What differences are there between the output this produces and the Full Documentation Experience? This overall looks good to me, I just want to make sure it'll properly include images and links; anything beyond that is nice but not necessary.

.github/workflows/build.yml
```diff
@@ -28,6 +28,8 @@ jobs:
       run: make --directory=./src dgdebug.exe dialogc.exe
     - name: make test
       run: make test
+    - name: build site with antora
```
hlship, line 31: I wonder if we could have the action check out the manual repo and do a full build there, then copy the results into this workspace?

dstelzer: That sounds ideal, if it's possible! I know it's possible to include a link to one repository inside another repository, which should get it checked out automatically, right?

dstelzer (approved):
It would be better to do a full build, but this is definitely better than nothing, and I think that's the right philosophy to have as we prepare for the first proper release. Approved.

bkirwi:
> What differences are there between the output this produces and the Full Documentation Experience?

You can download the generated HTML here, in the "artifacts" section:

https://github.com/Dialog-IF/dialog/actions/runs/13891371359?pr=73

It does support both links and images, but loses some of the custom styling.

Incidentally, you can find the output of any PR at a similar URI. (Weirdly hard to navigate to directly, but if you click on the actions tab at the top you get a list of recent runs to pick from.) Could be handy when reviewing major changes to documentation...

dstelzer:
Aha! Every day I learn about more Github features I'd never used before!

bkirwi:
I was a bit down on the idea of using the upstream `manual` repo for this documentation, since that includes potentially many versions... but then I (belatedly) realized I could use the local playbook from that repo directly, and it's already fairly close (or a little closer) to what we'd want.

I've switched this branch over to that approach. It's still slightly less than ideal - eg. it includes `Local` in the header, which isn't quite right for this case. But if folks want to make changes to how that works in the other repo, future PRs in this repo should pick that up.

bkirwi:
> Every day I learn about more Github features I'd never used before!

Yeah, I was only vaguely aware of this stuff before setting it up here. Pretty fussy but very useful when you get it working...

dstelzer:
Looking at the artifacts produced, this seems good to me! I just need to figure out how to sync my changes to the docs in this repo into the other, but if there's no fancy way to copy git commits from one repo to another, I can just do a diff and change them manually.

bkirwi:
Ah - the way hlship has this set up means you should't have to sync anything! This repo canonically contains the official docs themselves, and that one just has a pointer to this repo and some metadata. The reason for this otherwise rather baroque setup is it means that the other repo can have many pointers to different versions... so eg. someone using an old version of the compiler can go back and look at old versions of the docs. (But using the latest styling etc.)

Anyways thank you all for the review and the good suggestion... merging this in.

dstelzer:
Ah, right! I thought something had gone wrong for a moment because my changes to the documentation didn't appear in the compiled manual, but of course that's in #57 , which isn't in main yet. That all seems good then!

PR #74, opened 2025-03-17
FIX: update tests to handle the new meta-div behaviour

bkirwi:
The meta-div PR was merged without updating the necessary tests. This fixes up the tests to match!


dstelzer:
Oh shoot, good catch! I missed those.

dstelzer (approved):

bkirwi:
Thanks for the quick review!

Issue #24, opened 2025-01-22
Add (ambiguously plural $)


dstelzer:
Inform supports "ambiguously plural" objects, which set both IT and THEM when the object is noticed: a player who's carrying a PAIR OF BLUE JEANS should be able to both WEAR IT and WEAR THEM. This would be a minor change to the library and an easy improvement, with no real drawbacks; if a particular game doesn't use this feature, the compiler will optimize it out.

Issue #72, opened 2025-03-16
Gender-neutral fix causes Impossible Stairs tests to fail


dstelzer:
#40 has broken the Impossible Stairs test, because now the THEM pronoun gets redirected every time you enter a room with a person in it.

This seems like not ideal behavior. I think it would be better to only reset THEM when encountering an explicitly gender-neutral person; it's also possible now to have a person be both male or female and gender-neutral, to set multiple pronouns, and authors can use that to re-enable this behavior if desired.

If others agree, it's a one-line change to the library. But I want to hear other opinions before I touch it, especially from @averyhiebert who made the change in the first place. What do you think? Should we restore the original behavior, or update the test cases?

averyhiebert:
Changing back to the old behaviour for backwards compatibility is reasonable.

But I'm not sure that setting a character to both male/female and gender neutral simultaneously would work as a way to enable the new behaviour; I'm fairly sure it would mess up some verb conjugations (this could be fixed, but still...), in addition to being kinda conceptually weird/unintuitive.  Better in the long run to add a more narrow "can possibly be referred to as THEM, but with lower priority" predicate, as suggested previously elsewhere, but that doesn't need to be done right now.

dstelzer:
Fair! If you're okay with reverting this part of the library change, I'll incorporate that now, and we can come up with a more elaborate pronoun-setting system on another PR.

dstelzer:
Fixed in #43 .

PR #76, opened 2025-03-18
Add (ambiguously plural $) trait

dstelzer:
Mirroring Inform, an `(ambiguously plural $)` object sets both singular and plural pronouns. If it also has a trait like `(plural $)`, it will use plural pronouns; otherwise, it will use singular ones. Either way, both THEM and the appropriate one of HIM/HER/IT will be set to it.

This also allows THEM to be set to a person without making them gender-neutral, mitigating the issues created by fixing #72 .

### DIFF
--- a/stdlib.dg
+++ b/stdlib.dg
@@ -327,6 +327,7 @@
 
 %% These are typically static traits:
 
+(ambiguously plural $)	(fail)
 (an $)			(fail)
 (door $)		(fail)
 (edible $)		(fail)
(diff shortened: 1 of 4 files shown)


dstelzer:
This resolves #24 .

dstelzer:
@averyhiebert I'd also like to check if this works as a way to set THEM to a person without making them gender-neutral, since I ended up reverting that part of your earlier PR.

Now if you make a `(male $)` or `(female $)` object also `(ambiguously plural $)`, then they'll be referred to as "he" or "she" in text, and set the HIM/HER pronouns, but also set the THEM pronoun.

Does this work for your use case? If you want this to be true for all NPCs, you can add an implication that makes all `(animate $)` things also `(ambiguously plural $)`.

hlship (commented):
This looks good to me, but surely we have a mechanism to test this?

dstelzer:
Good point! I added a test case for this to the regression tests in new_ambigplural.dg. So that will now be automatically tested along with the rest.

hlship (approved):

PR #77, opened 2025-03-18
Implement status bar and div size controls

dstelzer:
Apparently I never actually PR'd this branch! My bad!


dstelzer:
See #78 for discussion and the rationale behind this.

dstelzer:
I've brought some of the regression tests into this branch to make sure nothing breaks; I had to delete the ones that test bugs that weren't yet fixed when this branch was first made, but those should get integrated when we merge this.

I also tried to add some regression tests for these new abilities. There's a new test for the div size measuring, but none for the status bar resizing, because dgdebug doesn't support status bars at all and dfrotz ignores resizing (it just prints everything from the status bar in the main window), so it can't really be tested automatically. The `.dg_` files in `test/library/simple` let us test it by hand if we want to.

This will be a good use case for remglk, once we have that all set up!

Issue #84, opened 2025-04-24
Fix all warnings under GCC


dstelzer:
Currently, compiling dialogc and/or dgdebug with GCC produces various warnings. They don't actually stop the program from compiling, but it would be nice to fix them just in case.

dstelzer:
I'm pretty sure one of these is actually a bug.

From eval.c:

```
	es->choice = u->choice;
	es->stopchoice = u->stopchoice;

	es->env = u->env;
	etop = envtop(es);

	assert(u->choice < es->nalloc_choice);
	assert(etop <= es->nalloc_env);

	memcpy(es->choicestack, u->choicestack, (u->choice + 1) * sizeof(struct choice));
	memcpy(es->envstack, u->envstack, etop * sizeof(struct env));

	while(es->divsp--) o_end_box();
	es->divsp = u->divsp;
	memcpy(es->divstack, es->divstack, u->divsp * sizeof(uint16_t));
	for(i = 0; i < es->divsp; i++) {
		o_begin_box("box");
	}
```

Note this line:
```
memcpy(es->divstack, es->divstack, u->divsp * sizeof(uint16_t));
```

This is copying an array to itself, which is undefined behavior. I'm pretty sure it should be copying `u->divstack` to `es->divstack` instead, just like the surrouding lines are doing.

PR #85, opened 2025-04-24
Fix obscure bug in undo implementation for debugger

dstelzer:
https://github.com/Dialog-IF/dialog/issues/84#issuecomment-2828784826

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -492,7 +492,7 @@ static int eval_pop_undo(struct eval_state *es) {
 
 	while(es->divsp--) o_end_box();
 	es->divsp = u->divsp;
-	memcpy(es->divstack, es->divstack, u->divsp * sizeof(uint16_t));
+	memcpy(es->divstack, u->divstack, u->divsp * sizeof(uint16_t));
 	for(i = 0; i < es->divsp; i++) {
 		o_begin_box("box");
 	}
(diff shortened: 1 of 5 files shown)


dstelzer:
For the other warnings, I've either:
- Fixed the minor bugs that prompted them (e.g. `snprintf(dest, len, string)` instead of `snprintf(dest, len, "%s", string)`)
- Added checks to satisfy the GCC compiler, even if they're not really useful
- Turned off those warnings in the Makefile

In particular, `-Wunused-result` cannot be silenced by anything in the code, such as casting the unused value to `(void)`; I have no idea why this is, but the only way to avoid those errors is to add a bunch of boilerplate carting unused values around, or to turn off that error message. I think turning it off is the sane choice.

Similarly, `-Wformat-truncation` warns if it can't guarantee that something will fit into `snprintf` output. This means a flood of warnings from across the codebase, basically every time a `%d` is used to print an integer into a small buffer. The compiler can't be sure that (e.g.) the current date won't be 999999-999999-999999, and thus gives a warning, but in practice that's never going to be a problem. If such an atrocious date _did_ show up, all that would happen is getting truncated, no buffer overruns.

So I don't think there's any bad consequence for turning off those two specific warnings. The program now compiles cleanly.

src/eval.c
```diff
@@ -492,7 +492,7 @@ static int eval_pop_undo(struct eval_state *es) {
 
 	while(es->divsp--) o_end_box();
 	es->divsp = u->divsp;
-	memcpy(es->divstack, es->divstack, u->divsp * sizeof(uint16_t));
```
bkirwi, line 495: :grimacing: - nice catch!

bkirwi (approved):

Issue #54, opened 2025-03-01
Support Å-machine options strings


dstelzer:
On the Å-machine, an embedded resource consists of a URL, some alt text, and a null-terminated _option string_ that's currently always empty. Allowing Dialog programs to set the option string provides an easy way of expanding Å-machine interpreters without actually changing the specification.

Currently, the syntax of a resource definition is:

```
(define resource ID) LOCATION ; ALT TEXT
```

I propose expanding this to:

```
(define resource ID) LOCATION [: OPTIONS] ; ALT TEXT
```

dstelzer:
Hmm, a colon might not be ideal because it can appear inside a URL. What else would be a good separator? I don't want to use a semicolon because I want it to always be unambiguous whether something is an option string or alt text; that'll ensure backward-compatibility with existing code.

Spaces aren't valid inside URLs, so we could just use that?

dstelzer:
Since there's already a place to store this information in the compiler's internal data structures, it could also be used on Z-machine. I'm not sure _how_ exactly it would be used, since right now the Z-machine backend doesn't handle linking or embedding resources at all (it just prints the alt text), but one could imagine e.g. a sound resource with `beep 1` or `beep 2` in its option tag being compiled as a Z-machine bleep instead of showing the alt text.

PR #89, opened 2025-05-19
Support option strings

dstelzer:
The compiler has always had a facility to include "option strings" with its resources. This pull request gives users a way to specify those.

Currently, they'll be included in generated Å-machine files, but are ignored by all current interpreters.

See #54 for more.

### DIFF
--- a/src/backend_aa.c
+++ b/src/backend_aa.c
@@ -3393,7 +3393,9 @@ static void chunk_urls(FILE *f, struct program *prg) {
 		size = 2 + prg->nresource * 2;
 		for(i = 0; i < prg->nresource; i++) {
 			offset[i] = size;
-			size += 3 + strlen(prg->resources[i].url) + 2;
+			size += 3 + // Alt text pointer
+					strlen(prg->resources[i].url) + 1 + // URL string w/ terminator
+					strlen(prg->resources[i].options) + 1; // Option string w/ terminator
 		}
 		pad = chunkheader(f, "URLS", size);
 		putword(prg->nresource, f);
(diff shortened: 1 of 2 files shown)


dstelzer:
Sample usage:

```
(define resource @aa) Makefile, option 1, option 2; alt text here
(define resource @bb) contributing.md , option 1	  
(define resource @cc) .gitignore; alt text here

(program entry point)
	(embed resource @aa)
	(embed resource @bb)
	(embed resource @cc)
```

Output when compiling with `-vvv`:

```
Resource 0:URL "file:Makefile", local file "Makefile", stem "Makefile", options "option 1, option 2", alt "alt text here"
Resource 1:URL "file:contributing.md", local file "contributing.md", stem "contributing.md", options "option 1", alt "contributing.md"
Resource 2:URL "file:.gitignore", local file ".gitignore", stem ".gitignore", alt "alt text here"
```

hlship:
I'll give this a once-over soon. I haven't coded in C in decades.

manual/modules/lang/pages/io.adoc
```diff
@@ -544,6 +544,15 @@ or a local filename.
 Local files are copied into the `.aastory` file,
 and this is the recommended way to work with _feelies_ and embedded graphics.
 
+The _location_ may be followed by one or more _options_, demarcated by commas. For example:
```
bkirwi, line 547: This is a little strong, I think - we don't require that the options are separated by commas internally, just the leading comma to separate it out.

bkirwi (commented):
It may be a little late to ask this - which is fine - but did we consider using a space as the delimiter? IIUC URIs can contain commas but not internal spaces... and as for local filesystem paths all bets are unfortunately off. :/

dstelzer:
Not too late at all! My rationale was that filenames on a local filesystem are much more likely to contain spaces than commas, so mandating no commas is more reasonable (on the user) than mandating no spaces.

manual/modules/lang/pages/io.adoc
dstelzer: Fixed!

Issue #88, opened 2025-05-19
Include -dev in version number in compiled Z-code files


dstelzer:
There's an eight-byte block at the end of the header where compilers and interpreters can store some information, traditionally in ASCII. Inform stores, for example, `Info6.13`, while Dialog stores `\0Dia1a01`.

That first null byte is currently unused. I propose we instead store a `*` there if compiling with a -dev version; `R_COMPILERVERSION` could check for that `*` and print `-dev` after the version number if so.

PR #91, opened 2025-05-20
Include -dev in version numbers on the Z-machine backend

dstelzer:
There's a byte in the header at $38, right before the version string, that currently goes unused. This PR stores `*` there if it's a -dev version of the compiler, and the Z-machine runtime checks for that to add `-dev` to the end of the displayed compiler version.

In other words, this means we can now distinguish between 1a/01 and 1a/01-dev in compiled output.


dstelzer:
#88

bkirwi (approved):
Clever!

Issue #90, opened 2025-05-20
Ensure the Z-machine header flags are accurate


dstelzer:
Now that the Z-machine backend supports color, for example, it should set the appropriate header flags to tell this to the interpreter.

dstelzer:
Flags 1: initial values don't matter

Flags 2:
- Bit 3: game wants pictures
- Bit 4: game wants undo
- Bit 5: game wants mouse
- Bit 6: game wants colors
- Bit 7: game wants sound
- Bit 8: game wants menus

Flags 3: only matters in version 6

Previously, Dialog always set bit 4 of flags 2 (wants undo) and nothing else. For color support, we'll change this to bits 4 and 6. In other words, from $0010 to $0050.

PR #93, opened 2025-06-09
Set flag 0 in Flags 2 if color is needed

dstelzer:
This shouldn't actually cause problems anywhere, but strictly speaking, a certain bit in the Z-machine header should be set if a game wants to use colors. This patch sets that bit if (and only if) a style class used in the game has a value of color or background-color that is not "inherit" or "initial".

It would probably be simpler to just always set the bit, like Dialog does for the "wants undo" bit, but why not add a little complexity and do it the right way?


dstelzer:
Fixes #90

PR #118, opened 2026-01-06
Merging many changes from the ifcomp2025 branch individually

dstelzer:
This is a broad, overarching pull request that I can split individual features and merge requests from.


dstelzer:
I think the other major changes have all been merged separately now, so let's see if all these ones can be incorporated automatically!

Issue #112, opened 2025-12-20
Z-machine runtime uses set_true_color without checking Standard revision


dstelzer:
Older interpreters won't understand the `set_true_color` opcode, so the runtime should check the Standard version like it does for `check_unicode` and fall back to a built-in implementation if it's not supported. (In particular, iOS Frotz doesn't support `set_true_color` and crashes on an invalid opcode.)

PR #119, opened 2026-01-06
Add Standard revision check before Z_TRUECOLOR

dstelzer:
The Z_TRUECOLOR opcode won't exist on interpreters before Standard 1.1, and in particular this can cause iOS Frotz to crash. So, we need to check the Standard revision before calling that opcode.


dstelzer:
Fixes #112

PR #140, opened 2026-01-28
unit.dg and examples added

susan-davis:
This PR adds unit.dg, a simple unit testing framework, and an example of its use in testing a sample extension.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dgdebug
 dialogc
 dgdebug.exe
 dialogc.exe
+log.txt
 *.z5
 *.z8
 *.zblorb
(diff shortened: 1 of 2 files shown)


dstelzer:
This looks excellent! My only concern: should we make a separate directory where extension libraries live? If the only Dialog files in the root are stdlib, stddebug, and unit, that's not too bad, but if we start adding further libraries, it's going to get unwieldy quickly; we should probably choose our approach now before it becomes an issue.

susan-davis:
We could have a `lib` where the `.dg` files go, or a `contrib` for extensions. I left my `time.dg` extension in `test/unit`.

A related question: hypothetical `unit.dg` tests for `stdlib.dg` and `stddebug.dg` -- should those go in `test/unit/library` by analogy with `test/simple/library`, or should the `test` directory be organized differently? A subject for some future PR, I think.

PR #142, opened 2026-01-29
Make unit.dg report in green until a test fails, then red thereafter (requires 1a/01 or later)

susan-davis:
Many other test runners report output in green until a test fails, then red afterwards. Use the colour predicates in 1a/01 and later to do this in `unit.dg`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ dgdebug.exe
 dialogc.exe
 **/log.txt
 **/.DS_Store
+**/*~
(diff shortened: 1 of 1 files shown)


susan-davis:
`dgdebug` won't output using the defined CSS without some hacking, and I have work to do on my WIP. Closing this; someone can do this later if the behaviour is desired. (This was always just a "nice to have," and seemed like low-hanging fruit, which it isn't.)

Issue #127, opened 2026-01-08
Quick syntax reference lacks cross-references


dstelzer:
The "quick syntax reference" at the end of volume 1 of the manual lists all the special syntax and built-in predicates, but has no cross-references back to where that syntax is explained. The index in volume 2 has these cross-references, but only for the predicates, not for the special syntax.

We should either add cross-references to the quick reference, or include special syntax in the index.

dstelzer:
The latter seems easy enough to do, but we should probably do both, really.

PR #144, opened 2026-01-30
Improve indexes of both manual books

dstelzer:
This PR accomplishes two things:
- It adds hyperlinked cross-references to the syntax reference at the end of the language book
- It includes all "predicate-like" special syntax in the reference at the end of the library book

The end result is that you can look up any syntax in the language index and jump to the page where it's explained, or look up anything "predicate-like" (words in parentheses) in the library index and likewise jump to the explanation.

While we still can't search for things like `(if)` with Antora, this should help.

(#127)

### DIFF
--- a/manual/modules/lang/pages/beyondprg.adoc
+++ b/manual/modules/lang/pages/beyondprg.adoc
@@ -1,7 +1,6 @@
 = Chapter 9: Beyond the program
 
 [#metadata]
-
 == Story metadata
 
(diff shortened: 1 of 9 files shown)


bkirwi (approved):
Very nice!

I have not reviewed all the cross-linking in detail, but regardless this seems like a nice improvement. Thanks in particular for all the typo fixes.

PR #146, opened 2026-01-31
Add (nbsp) predicate for non-breaking spaces

dstelzer:
With this PR, `(nbsp)` joins the suite of spacing predicates (`(no space)`, `(space)`, `(line)`, `(par)`) , producing a non-breaking space. In the debugger, it outputs a space without activating the line-breaking machinery. On Z-machine, it prints a Unicode NBSP character (0xA0) if possible, and a normal space otherwise. On Å-machine, it calls a new opcode, EXT0:NBSP.

It interacts with the other spacing predicates as you might expect: it overrides `(no space)` and is overridden by all the others, including `(space)`.

The Z-machine and debugger versions seem to work, and you can check with the new test case added in this branch. On Å-machine, I have a new version of all the interpreters (Node, web, C64) ready, but I'm having trouble testing them properly. We should figure out a better way to test that.

This also requires bumping the Å-machine version up from 0.5 to 1.0, since we added a new opcode. (Switching to major version 1 for the same reason we jumped from Dialog 0m/03 to 1a/01.)


manual/modules/lang/pages/execution.adoc
```diff
@@ -153,6 +153,15 @@ This is not broken up.
...
 ```
 
+[#nbsp]
+=== The gory details
+
+When you call one of the spacing predicates, `(no space)`, `(nbsp)`, `(space)`, `(line)`, and `(par)`, Dialog doesn’t display anything immediately. Instead, the next time a word needs to be printed, Dialog inserts _the largest space that’s been requested since the last word_. This means `(no space) (space) (nbsp) (no space) (nbsp)` is equivalent to a single `(space)`, and `(par) (line) (line) (par) (space) (space)` is equivalent to a single `(par)`.
+
+The same goes for xref:io.adoc#div[div margins], which will be introduced in a later chapter. A `(par)` before a div with a one-line margin will still produce only a single blank line.
+
+(The rarely-needed `(nbsp)` creates a “non-breaking space”, which looks identical to a space, but will not allow a line break at that location. It’s mainly used between numbers and units, like “100 km” or “section 5”, and before certain punctuation marks in French. If an interpreter cannot provide this, it will produce a normal space.)
```
Natrium729: Maybe be a bit more generic? Instead of saying “and before certain punctuation marks in French”, maybe something like “and in other languages, such as before certain punctuation marks in French” (because I’m sure that doesn’t concern only French). (Also technically it sometimes comes *after* the punctuation, such as opening quotation marks. Anyway.)

Also, “section 5” is not really an example of “between numbers and units”, but I’m nitpicking.

dstelzer:
Dgdebug is weirdly fancy about word breaking: *if* it can detect that it's running in a terminal, and *if* it can measure that terminal's width, it does its own special line-breaking code that tries to only break between words. So if those conditions are met, it'll do a proper non-breaking space. (If they're not met, which is often, it'll just print a space.)

For Å-machine interpreters, there'll be new terps soon that implement this, but that involves a separate pull request on a separate repo that's not worth thinking about now.

Natrium729:
> Dgdebug is weirdly fancy about word breaking: if it can detect that it's running in a terminal, and if it can measure that terminal's width, it does its own special line-breaking code that tries to only break between words.

Oh, OK. Since I checked by resizing the terminal window to see if the punctuation wrapped, it’s normal it wouldn’t work. (I’d still make it print a real non-breaking space for the sake of copy-pasting and such, but really not a big deal either since it’s just used during development.)

Otherwise, I’d say everything’s good! (And anyway, I’ll report if ever I encounter an issue.)

PR #147, opened 2026-02-01
Add (quit $) and (quit minus $) builtins to dgdebug

susan-davis:
Add `(quit $)` and `(quit minus $)` builtins to `dgdebug`, to let `make` fail the build if `unit.dg` tests fail in `dgdebug`. Silently ignore the argument in `dialogc`.

### DIFF
--- a/test/simple/language/Makefile
+++ b/test/simple/language/Makefile
@@ -114,9 +114,9 @@ clean:
 	rm -rf *.out *.z.gold *.aa.gold *.d.gold *.z.in *.aa.in *.d.in *.z5 *.aastory
 
 test-quit-0.test: $(DGDEBUG)
-	$(DEBUG) test-quit-0.dg || exit 1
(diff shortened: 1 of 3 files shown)


dstelzer:
I'm definitely in favor of `(quit $)`, but is `(quit minus $)` needed? I thought exit statuses were supposed to be between 0 and 255 in POSIX.

dstelzer:
Side note: it looks like `(quit $)` didn't get added to the built-in predicate reference or the full syntax reference. It's no big deal to add it now (I'll make a new PR for that), but for the future, I'd recommend leaving PRs open for a few days before merging them, so that they can get other eyes on them. (I didn't look deeply at this one because I didn't realize it was ready to merge.)

Personally, I try to leave all PRs open for a week or until someone else signs off on them, and major changes open for a month or until someone else signs off on them.

Issue #59, opened 2025-03-06
Allow (status bar $ with height 0) to remove all status areas


dstelzer:
Currently, (status bar $ with height $) has a minimum of 1em. It would be nice to allow removing the status bar entirely by specifying a height of 0.

dstelzer:
Or similarly, using a style class with a height of 0em.

PR #149, opened 2026-02-01
Add (clear status bars) predicate

dstelzer:
Previously, it was possible to clear the entire screen with `(clear all)`, or clear only the main window with `(clear)`. There was no option to clear only the status areas.

Now there is: `(clear status bars)`. It does what it sounds like. On Å-machine this requires a new opcode, EXT0:CLEAR_STATUS. Since interpreters can legally ignore any of the clearing builtins, implementation won't be hard.

### DIFF
--- a/manual/antora.yml
+++ b/manual/antora.yml
@@ -1,6 +1,6 @@
 name: dialog
 title: Dialog
-version: 1b01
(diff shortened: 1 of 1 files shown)


dstelzer:
Fixes #59

Issue #107, opened 2025-08-24
What happens when the Z-machine runs out of registers?


dstelzer:
Dialog is generally quite efficient at allocating Z-machine registers, so I've never seen it run out. But as best I can tell, the compiler doesn't actually have any logic to check this. It'll just wrap around to zero and start reusing earlier ones without warning.

We should figure out what the current compiler actually does, and if this needs to be updated.

PR #151, opened 2026-02-02
Report usage of Z-machine registers

dstelzer:
All versions of the Z-machine have 240 registers that can be accessed from anywhere in the program. Dialog uses these for various purposes, but previously didn't check if they ran out.

Now it does. It also reports how many are used for various purposes if compiling with `-vv`.

Miss Gosling:
```
Debug: Registers used: 113 of 240 (47%)
Debug:                 61 internal, 19 temp, 33 global
```

Impossible Stairs:
```
Debug: Registers used: 122 of 240 (50%)
Debug:                 61 internal, 32 temp, 29 global
```

So hopefully nobody is running out any time soon!

### DIFF
--- a/src/backend_z.c
+++ b/src/backend_z.c
@@ -4473,6 +4473,11 @@ void backend_z(
 	nglobal = (REG_X - 0x10) + max_temp;
 	user_global_base = nglobal;
 	nglobal += next_user_global;
+	
+	if(nglobal > 240) {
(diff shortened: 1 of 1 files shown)


dstelzer:
Out of the various uses, userglobals (that is, `(global variable $)` variables) would be the easiest to shift into random bits of RAM if necessary. Hopefully it won't be necessary.

dstelzer:
#107

PR #154, opened 2026-02-03
add 'make tidy', 'make distclean', and 'make uninstall' targets

susan-davis:
Add the following `make` targets:
- `make tidy`: delete object and other generated files, but leave `src/dialogc` and `src/dgdebug` alone
- `make uninstall`: undo `make install`
- `make distclean`: get rid of everything that wasn't in the tree when you did a `git clone` or `git checkout`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,10 @@ tidy:
 	$(MAKE) -C src tidy
 	$(MAKE) -C test/gosling clean
 	$(MAKE) -C test/impossible clean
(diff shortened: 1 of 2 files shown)


susan-davis:
Should we also install `stdlib.dg`, `stddebug.dg`, and `unit.dg` to `/usr/local/lib` or some other user-specified libdir? That would let someone building from source nuke their source tree after installation.

Issue #157, opened 2026-02-04
dgdebug: improvements for automation -- dynamic state output


hlship:
I'm interested in capturing the dynamic state (global flags, global vars, per-object flags, per-object vars) after each command is executed.

My current approach is to run the command `@dynamic` after each command, and then parse the result.
 
Parsing is complicated by word wrapping; not insurmountable, but I'm concerned about edge cases I haven't found examples for.

But of course, I had to wonder: could I make this someone elses' problem?

I would love to see additional command line flags that would enable JSON output when commands execute.  The JSON could have one key for main command output, and additional keys (if enabled) for all the dynamic data.

Here's a potential example (heavily abbreviated):

```
{"output": "You are standing in a clearing .... "
  "global_flags": {"(inhibiting next tick)": "off",
                             "(head noun is required)": "off (changed)"},
   "per_object_flags: {"($ is closed)": ["#drawer", "#gas-can"]},
   "global_vars":  {"($ is visited)",  "#crumbling-concrete #roof #storage-room #middle-of-nowhere"},
   "per_object_vars": {"($ has parent $)": [["#flashlight" "#knock"],
                                                                         ["#cigarette" "#coyote"]]}}
```

PR #160, opened 2026-02-04
Disable word wrapping when --width -1

dstelzer:
Exactly what it says on the tin: this PR offers a way to disable word wrapping in the debugger, by setting `force_width` (`--width`, `-w`) negative. This was requested in #157 for when the output is piped into other programs.

Previously the only way to do this was to specify an absurd width, but this led to allocating a huge buffer for the word wrapper. The new solution is better.


hlship (approved):
Though I have to wonder what the behavior of `-w 0` would be.

PR #161, opened 2026-02-04
Support (uppercase) for extended ZSCII characters

dstelzer:
The compiler will now add uppercase versions of any extended ZSCII chars to the inventory, and a new R_EXT_UPPER will apply them when `(uppercase)` is used.


dstelzer:
The question is, should they be added to ZSCII, or should they be stored in the table as Unicode values for `@print_unicode`? This takes up a bit more space in the repertoire (and means a translation table will be needed even for the standard inventory, because of `ÿ`), but it's also more compatible with older interpreters.

PR #164, opened 2026-02-18
Support (current div width/height $) on Å-machine backend

dstelzer:
The new logic for these is:
- Call VM_INFO to get the information
- If the VM returns $4000 (that is, the number 0), fail
- Otherwise, return that number

Note: this requires an updated Å-machine interpreter. The tests will fail until the equivalent PR for the Å-machine interpreter repository is merged.


dstelzer:
Do you have the latest version of the Å-machine Node interpreter? The latest version from the main branch of https://github.com/Dialog-IF/aamachine/ should (1) support those VM_INFO codes, and (2) not crash on an unrecognized VM_INFO for the future.

PR #188, opened 2026-03-07
Add --tag-lines option to dgdebug

dstelzer:
This option is mainly useful for automation; it prepends output lines with `  `, line input with `> `, and key input with `) `. This makes it possible to tell what the program is waiting for.


dstelzer:
Good point! Check out test/simple/debugger/linetags.gold and compare against test/simple/debugger/nolinetags.gold in the new test cases. That runs the same program, once with line tags (and no header), then again with default settings.

PR #201, opened 2026-03-15
Add (tagged name $) predicate

### LINKED ISSUE #139: Before/after printing the name of a thing
Inform 7 has a very convenient, versatile way to alter the names of things: "printing the name of" something is an activity, so it can have "before", "for", and "after" rules.

I don't think Dialog needs something that big. But in two separate projects now I've found it very annoying to stick a tag onto the start or end of a name, not for one object, but for an entire class of them.

Right now, that's very difficult, because the standard library assumes that `(name $)` is both the predicate you define to give a thing's name, and the predicate you call to print that name. Splitting those into separate predicates would give authors more versatility to adjust the process, and the optimizer should keep this efficient in the most common case.

For example, it might look like:
```
(full name $Obj) (name $Obj)

(the (your $Obj)) your (full name $Obj)
(the (object $Obj)) the (full name $Obj)
...
```

Now `(name $)` is the predicate that authors define in their code, same as before, while `(full name $)` is the predicate that the library uses to print back those names. By default, the two are equivalent, but you could override that first rule for your own purposes:

```
(full name (enchanted $Obj)) enchanted (name $Obj)
(an (enchanted $Obj))
```

Or, if you really needed the full versatility of the Inform version:

```
(full name $Obj)
    (exhaust) *(before name $Obj)
    (name $Obj)
    (exhaust) *(after name $Obj)
```

dstelzer:
This will most often go unused, but allows an equivalent of Inform's "before printing the name of" and "after printing the name of" if needed. If it's not used, the compiler will optimize it out completely.


dstelzer:
That is correct! This is equivalent to just defining an `(actual name $)` predicate for myself.

My rationale for putting it into the library is that Dialog authors are very used to writing `(name $)` every time they define a new object. Having the library use the awkward predicate name instead of the author minimizes the chance of mistakes there. And if the author doesn't use the feature, it gets optimized to nothing—compare features like fungible objects and choice mode that go unused in most projects.

(This has come up twice for me specifically: once to put the spell adjectives on objects in Wise-Woman's Dog, a second time to put liquid names in front of containers in a TALJ project. Once, I was happy to hand-roll; twice, I started wondering if the library should provide the hook for me.)

If others disagree, though, I'm happy to be outvoted on it.

dstelzer:
[The poll was absolutely neck and neck](https://intfiction.org/t/dialog-wishlist/70278/434?u=draconis), but "add `(tagged name $)`" won by a hair, so I'm going to merge this in. If someone wants to remove it in the future, though, I won't stand in their way.

PR #203, opened 2026-03-15
Implement (body style $) and (reset body style) predicates

### LINKED ISSUE #200: Add global style predicate
Dialog currently has no way to set the _overall_ CSS of the game, outside of any divs and spans. This would be nice for things like the background color changes in _The Van der Nagel Papyrus_ or the global text color changes in _Photopia_. It would also allow authors to embed an overall color scheme in the game itself, rather than editing the Å-machine web interpreter's CSS.

I'm imagining something like `(global style $)` that applies a style class to the entire screen (minus any status bars). For simplicity, this predicate would be illegal inside a div or span; whether it applies to existing text is undefined (Z-machine no, web yes), so it's safest to `(clear)` immediately after.

On the Å-machine, this would require a new opcode; I propose \$ED, which is in the "output control" range (\$6X and \$EX) and not currently defined. On the Z-machine, it would just involve setting some registers.

https://intfiction.org/t/dialog-wishlist/70278/388?u=draconis

dstelzer:
See #200 for rationale


dstelzer:
The Å-machine PR has gone through, and thus the tests now pass again! Un-drafted this as a result.

susan-davis:
Was this meant to work in the debugger? Colours do, but this doesn't.

dstelzer:
It currently doesn't, because with only the eight basic colors implemented, I figured the results wouldn't look good. But it's trivial to add if desired.

PR #206, opened 2026-03-16
Use Dialog-specific Lunr extension fork for issue #148

### LINKED ISSUE #148: Manual: the search index doesn't work well with dialog predicates
In forums: https://intfiction.org/t/how-should-extensions-non-stdlib-libraries-be-organized/78837/3?u=hlship

> If the manual is a factor in the decision I think you should delay deciding how to organise anything until it is proven that Antora can actually produce a usable search index. The sites that it builds look nice, but currently you can’t search for any Dialog syntax because the index appears to have been created with punctuation stripped out - it is a fairly fundamental issue that you can’t search for language syntax like (if) because it isn’t going to be much use to search for if. It wouldn’t be great if the same limitation gets inherited by anyone documenting or using an extension.

I have little doubt that Antora's search index can be configured to work better; I don't know the specific however. It is something to track.

Randozart:
Installs the forked extension with custom trimmer preserving parentheses and @ symbols for Dialog syntax search.


dstelzer:
@Randozart Any success on this? No pressure intended; I'm just checking on the status of outstanding pull requests as we head toward a release.

Randozart:
@dstelzer Hey, sorry for the delayed response. Had a chaotic week behind me, and to top it off, I finally switched to Linux. I have the fix in a folder on my PC, so I'll perform a few sanity tests this week, then push it and test it here.

dstelzer:
Wonderful! No pressure intended; we're all doing this in our free time, after all. I'm just trying to figure out when a feasible next release would be.

PR #213, opened 2026-03-17
Document testing in Dialog

### LINKED ISSUE #173: Manual: add chapter on testing
With `unit.dg` now part of the distro, there should be a manual chapter covering testing. It should also cover the "run a game with a walkthrough, and diff against expected output" pattern, and possibly mention `dgt` and its skein.

susan-davis:
Add a Chapter 12 to the library section of the Dialog manual, explaining how to write and run `unit.dg` tests, and possibly other topics.


susan-davis:
There's more that I want to write, but the amount that's written so far is useful, and is better gotten out there.

manual/modules/lib/pages/testing.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+(interface (max $<X $<Y $>Max))
+
+(interface (min $<X $<Y $>Min))
+
+(max $X $Y $Result)
+     (if) ($X > $Y) (then)
+     	  ($Result = $X)
+     (else)
+	  ($Result = $Y)
```
dstelzer: The indentation on the \$X and \$Y lines looks different in Github's interface—it only uses two spaces for a tab, so for consistency I think the code samples should either use all spaces or all tabs.

manual/modules/lib/pages/testing.adoc
```diff
@@ -0,0 +1,369 @@
...
+Attempting 1 test.
+Testing #max-first-is-bigger: Passed!
+1 test passed successfully.
+$
+----
+
+Let's look at some of the details of our first draft of
+`minmax-tests.dg`. First of all, by convention, we put a comment in the
+first line of the file with the command line that we need to run to
+successfully run our tests under the debugger. We run unit tests with
```
dstelzer, line 87: The word "run" three times in these two lines feels a bit off, but if you don't want to reword it at this point, it's not a big deal.

manual/modules/lib/pages/testing.adoc
```diff
@@ -0,0 +1,369 @@
...
+[#unittests]
+== Unit Tests With `unit.dg`
+
+Included in the Dialog library is `unit.dg`, a library that allows you
+to write unit tests for your code in Dialog. See the
+xref:ROOT:software.adoc[Software Page] for more information on where
+to find `unit.dg`.
+
+Unit tests are short, simple tests that exercise small pieces of your
+code. They can save quite a bit of debugging, and even more helpully,
```
dstelzer: *helpfully

manual/modules/lib/pages/testing.adoc
```diff
@@ -0,0 +1,369 @@
...
+will let you know if you've accidentally broken something unrelated
+when you make a change to some other part of your game, if you run
```
dstelzer: I think the ", if…" part of this sentence isn't needed.

manual/modules/lib/pages/testing.adoc
```diff
@@ -0,0 +1,369 @@
+= Chapter 12: Testing and Debugging
+
+Interactive fiction games such as those that you'll write with Dialog
+require a lot of testing to get right. Dialog provides several
+facilites to help you test your code, and a full-featured debugger to
+work out why your code isn't behaving as expected.
+
+[#unittests]
+== Unit Tests With `unit.dg`
```
dstelzer, line 9: Right now, this chapter has a really good writeup of how to unit-test extensions, but no details on how to unit-test IF things (like making sure an action works properly). That's not necessarily a bad thing—an extension is probably where unit tests are most helpful—but I would mention it at the top of this chapter somewhere. "These unit tests are especially valuable for new libraries and extensions."

That caveat can then be removed if/when more details on testing the IF aspects are added (e.g. `(try [take #non-item])` and asserting that the object didn't move).

manual/modules/lib/pages/testing.adoc
```diff
@@ -0,0 +1,369 @@
...
+(run *)
+    (8000 plus 9000 into 10000 max 10000)
+
+#16383-plus-16383-max-16383
+(test *)
+(run *)
+    (16383 plus 16383 into 16383 max 16383)
+----
+
+But when we run this, we get the following:
```
dstelzer, line 288: I like this example!

dstelzer (changes requested):

dstelzer (approved):

dstelzer:
Looks good to me!

PR #220, opened 2026-03-23
Add ability to force height in debugger

### LINKED ISSUE #115: Debugging commands to change console dimensions
[Source](https://intfiction.org/t/dgdebug-for-automation/73518/8?u=draconis)

Commands to change the "height" and "width" of the console would be useful for multiple purposes. It could turn off paging before executing a `@replay`, for example, or manually set a height and width even when pipes are involved (which normally disables paging).

dstelzer:
New `--height` option acts like `--width` but for terminal height. Useful for disabling `[more]` prompts.

### DIFF
--- a/manual/modules/ROOT/pages/software.adoc
+++ b/manual/modules/ROOT/pages/software.adoc
@@ -300,6 +300,14 @@ Shows the current state of all dynamic predicates (global flags, global variable
 per-object flags, and per-object variables). For global flags, it also shows
 whether they’ve been changed via `(now)`.
 
+@more::
+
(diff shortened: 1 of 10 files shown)


dstelzer:
`@more` and `@nomore` debugging commands set --height to 0 and -1 respectively.

dstelzer:
`--unit-test` (or `-u`) command-line option is equivalent to `--no-warn-not-topic --quit --height=-1`.

susan-davis (approved):
Great!

susan-davis:
Maybe in some future PR, we could add `--no-header` to `-u`?

PR #221, opened 2026-03-24
Add color support to the debugger

### LINKED ISSUE #182: Debugger does not support colored text
It sounds like this would be desirable for the unit tests. It would be a bit annoying to add to the debugger, but far from impossible.

dstelzer:
The implementation is very basic at the moment: only supports eight named colors plus `initial` and `inherit`, and depending on the terminal may put the wrong color on the line following a div. But it allows printing text in green or red when unit tests pass or fail, and that's the main thing I expect people to use it for.

### DIFF
--- a/src/term_tty.c
+++ b/src/term_tty.c
@@ -141,7 +141,7 @@ static void reset_bgcolor() { // Reset the background color after a newline
 }
 
 int term_sendlf() {
-	printf("\033[49m"); // Always reset background color before scrolling, to avoid weird behavior with the new line produced
+	if(isatty(1)) printf("\033[49m"); // Always reset background color before scrolling, to avoid weird behavior with the new line produced
 	fputc('\n', stdout);
 	reset_bgcolor();
 	if(io_tag_lines) printf("  ");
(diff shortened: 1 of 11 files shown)


src/term_tty.c
```diff
@@ -110,10 +111,29 @@ void term_effectstyle(int style) {
...
 	}
 	termstyle = style;
 }
 
+void term_colors(int fg, int bg) { // OCOLOR_* = ANSI escape color (0-7 or 9)
+	int cat;
+	if(fg != termfg && isatty(1)) {
+		assert(fg != OCOLOR_INHERIT); // INHERIT should never get this far - we should only be sent actual colors at this stage
+		cat = fg ? 9 : 3; // Use code 9X (bright colors) if not black; for black, use 3X (dim colors)
+		printf("\033[%d%dm", cat, fg);
```
susan-davis, line 126: If this had been 35 years ago, I'd have had something to say about using direct escape sequences with an assumption that it's an ANSI terminal, rather than using curses/terminfo, but essentially no one these days is using an actual dumb terminal, and those few hobbyists who are aren't using colour ones, so just assuming ANSI is fine.

susan-davis (approved):
This should let me move forward with adding the corresponding support to `unit.dg`.

src/term_tty.c
dstelzer: Yeah, the debugger has already been assuming ANSI for italics and bold and such, and I think that's a pretty safe bet at this point.

PR #224, opened 2026-03-27
Fix Bocfel color bug

dstelzer:
[Bocfel (or another popular terp?) doesn't properly reset colors when restarting.](https://intfiction.org/t/what-controls-the-overall-background-color-on-z-machine/79427/9?u=draconis) So this PR resets the colors to default right before executing a `@restart` opcode.


dstelzer:
I think I screwed up that merge. Oh well.

PR #225, opened 2026-03-29
support colours in unit.dg

susan-davis:
Make `unit.dg` output in green until a test fails, and red thereafter. Include a switch to turn off the colours if desired.

For some reason, (body style $) isn't working in the debugger, which made this uglier than it had to be.


dstelzer:
Please give me at least a day before deleting the branch if you request my review XD I was in the middle of testing this when you merged it

Looks good to me, though; no objection to the merge itself

PR #226, opened 2026-03-29
Implement (body style $), (reset body style), (current div height $) in debugger

dstelzer:
These don't work _great_ in the debugger, but they do work to the specification, so there's no reason not to allow them.

Note, however, that the debugger does _not_ support `(clear)`, which limits the effectiveness of body styles.

image

### DIFF
--- a/src/dumb_output.c
+++ b/src/dumb_output.c
@@ -5,6 +5,7 @@
 #include "report.h"
 
 static int force_width;
+static int force_height;
 
 static void bail_out() {
 	report(LVL_ERR, 0, "Output is not allowed during initial value evaluation.");
@@ -37,6 +38,7 @@ void o_post_input(int external_lf)		{ bail_out(); }
 
 void o_reset(int force_w, int force_h, int quirks) {
 	force_width = force_w;
+	force_height = force_h;
 }
 
 void o_leave_all() {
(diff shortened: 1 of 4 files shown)


dstelzer:
@susan-davis Check this out and let me know if it works for your purposes. If you just want to change the color of the text output in an imperative way (rather than launching new divs/spans), it should be fine.

susan-davis:
@dstelzer If `(body style $)` threw a warning (suppressed by `-u`) rather than an error if the parameter were of a type other than a dictionary word, so that you could check the current body style with `(body style $Style)`, what would that break? (It's a nice-to-have, not an absolute need, if it would be like kicking dead whales down the beach to implement that.)

susan-davis (approved):

dstelzer:
Unfortunately not possible with the current implementation—the current body style isn't actually saved anywhere. Calling the predicate adjusts a bunch of internal variables, then succeeds, so which dictionary word exactly you called it with is lost.

This could be changed in the debugger and Å-machine, but that wouldn't work on Z-machine, and I worry about breaking compatibility between the implementations. For now, I would recommend writing a wrapper around `(body style $)` that also sets a flag you can check. That will work equally on all backends.

PR #228, opened 2026-03-30
List failing tests on exit from unit.dg

susan-davis:
Have `unit.dg` summarize the failing tests at the end of the run, with a mechanism for turning the behaviour off.


susan-davis:
Branched off the wrong branch.

PR #230, opened 2026-04-01
don't print header when running unit tests

susan-davis:
There's no reason to print the `dgdebug` banner when running a suite of unit tests, and the instruction in it to type "help" at the debugger prompt is inappropriate. Make `-u` also include `--no-header`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ tidy:
 clean: tidy
 	$(MAKE) -C src clean
 
-install: test
+install: src/dialogc src/dgdebug
 	$(MAKE) -C src install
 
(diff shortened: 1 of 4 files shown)


dstelzer:
Implementation looks good to me! The documentation should be updated to match, though—I think the only places `-u` is documented is in the `dgdebug --help` output and in `readme.txt`.

susan-davis:
Suppressing the header also has the nice side effect that warnings stand out visually. When headers were being printed at the start of every test file, the warnings blended into that noise.

dstelzer (approved):

Issue #234, opened 2026-04-11
@nomore command leaks color and removes some output lines


morganwillcock:
When trying to use the `@nomore` command, the color of the next line appears to be wrong and some empty lines are not displayed.

Here is a screenshot of the problem when using `@replay`, although just playing the game normally will also show the same problem:
<img width="802" height="857" alt="Image" src="https://github.com/user-attachments/assets/9bda76dc-0e7c-4ad6-b55e-ab4670f9eb80" />

Using `@more` also changes the color of the next line.

If I use the latest version of the standard library I get the same result (minus the warnings). I am testing the version of `dgdebug` that is built from commit fa0234f67850abc16407c74cce61a33dae92e36e.

dstelzer:
Good catch! Give #235 a shot and see if that fixes it. (It appears to on my machine.)

morganwillcock:
It fixes the color of the next line but the empty lines of output are still missing.

dstelzer:
Found it! When printing multiple blank lines in a row (for the top or bottom margin of a div), it checked if that number was larger than the size of the window, and constrained it if so. But `@nomore` set the size of the window to -1, so _every_ run of blank lines was larger than the size of the window.

Try it now (same branch).

morganwillcock:
Did you push the changes?

dstelzer:
Oops, looks like a merge went bad! Fixed.

morganwillcock:
That seems to fix everything. Thanks!

PR #235, opened 2026-04-11
Ensure boxes are closed before calling o_reset

dstelzer:
An oversight in the `@more` and `@nomore` commands led to colors bleeding onto the next line. This PR should fix that by closing the debugger box before resetting the terminal.

### DIFF
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -1147,12 +1147,16 @@ static void cmd_restore(struct debugger *dbg) {
 
 static void cmd_more(struct debugger *dbg) {
 	force_height = 0;
+	o_end_box();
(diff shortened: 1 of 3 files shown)


dstelzer:
Fixes #234

dstelzer:
Looks like I'm using this branch to fix some other bugs in the output system as well:
- With a forced height of -1, runs of newlines (e.g. in div margins) would not be displayed
- If the terminal can't report screen dimensions, assume 80×∞ instead of 0×0 (which causes problems for the word wrapper)

dstelzer:
I've moved the word wrapping stuff into #239 instead, so gotta merge that one first.

Issue #222, opened 2026-03-24
Debugger color support is not great


dstelzer:
Once #221 is merged, the debugger will support basic colors, but not much beyond that. In particular, after a div with a background color, the next line might get painted in that color depending on the terminal; and only named colors are supported, even though there are ANSI codes for full 24-bit color.

This is a low priority for me, but I figure it should go in the issue tracker regardless. If someone else wants to fix it, be my guest!

dstelzer:
Fixed the line issue. Still no support for 24-bit colors.

susan-davis:
Weirdly, the named "green" that it uses is a different and less-legible shade of green than the one that `git` uses. I kept the summary on the top and bottom lines in the default colours, and any output that a test itself produces will also be in default, not red or green.

dstelzer:
Hmm. Let's find out what ANSI code Git is using!

dstelzer:
Looks like Git's red is [31m and green is [32m. That is, it's using the dim colors, not the bright colors.

dstelzer:
Wow, yeah, both the dim and the bright colors look _bad_ on my terminal: [link](https://intfiction.org/t/is-it-better-to-use-30m-colors-or-90m-colors/79571/17?u=draconis)

I've asked on the forum about what to do.

PR #237, opened 2026-04-11
Switch to user-suggested color scheme

dstelzer:
This PR replaces the default ANSI colors with the ones from https://intfiction.org/t/is-it-better-to-use-30m-colors-or-90m-colors/79571/48?u=draconis .


dstelzer:
It looks good on my Ubuntu terminal, but I'm going to ping in @susan-davis to make sure it works on Mac as well.

This will resolve the main concerns from #222 , though it still won't allow providing a full 24-bit color in hex.

dstelzer:
I'm also going to ping in @hlship , mainly as a warning. This is going to make it harder to recognize specific colors in dgt, _but_ dialogc will now be producing full 24-bit color values, so if you parse those out they'll be even easier to display than before.

dstelzer:
Specifically, this is what will be happening now:
```
static int32_t ansi_to_24bit_color[] = {
	0x262626, // black
	0xdf2050, // red
	0x18aa49, // green
	0xb5a926, // yellow
	0x6361ea, // blue
	0xbf1cbf, // magenta
	0x21baba, // cyan
	0xf2f2f2, // white
};
// ...
printf("\033[3");
if(fg == OCOLOR_INITIAL) {
	printf("9m");
} else {
	color = ansi_to_24bit_color[fg];
	printf("8;2;%d;%d;%dm", color>>16, (color>>8)&0xff, color&0xff);
}
```

And same with 4 in place of 3 for background color. So the new codes will look like `\e[38;2;255;255;255m` where those three "255"s are the red, green, and blue components as decimal numbers from 0 to 255. Colors are reset with `\e[39m` as before.

For now, those eight colors are the only ones supported. Future versions might let users specify full 24-bit colors too.

dstelzer:
…hmm. Though, if it would help dgt, I could also output an old-school ANSI escape sequence before switching to the new color. That would probably be easier for dgt to parse.

For example, I could output `\e[30m\e[38;2;...m` when setting `color: black;`. First, set black text using the old-school sequence; then, set the exact 24-bit color code.

Would this be useful for dgt? Or is it just a waste of bytes?

Issue #238, opened 2026-04-11
Add "dumb" version of dgdebug on Windows


dstelzer:
The Windows terminal is not very full-featured, but the Glk version of dgdebug doesn't work with dgt. So for dgt purposes, it would be nice to have a basic terminal version of dgdebug for Windows that doesn't use Glk.

PR #239, opened 2026-04-11
Add stripped-down TTY implementation for Windows

dstelzer:
See #238 for rationale


dstelzer:
It runs in WINE for the moment, which is a good sign. Keeping two versions of term_tty.c in sync will be a pain, but only the very basic functionality is needed here.

dstelzer:
A bunch of `#ifdef`s now mean that we don't need to keep two versions of the file in sync. Much better.

PR #242, opened 2026-04-15
Work toward feature parity between Windows TTY and Mac/Linux

### LINKED ISSUE #241: Consider renaming Windows debuggers
At the moment `dgdialog.exe` is a graphical progam and `dgdialog_tty.exe` is the terminal/command-line version. I think it makes sense to match the naming with the other platforms so that graphical programs are named in a way that is different to the non-graphical versions, e.g. `dgdebug` and `dgdebug.exe` are the non-graphical applications, while `dgdebugw.exe` (or similar name) is the graphical version.

Also, the "_tty" in the name is a little misleading because it can work without a TTY.

dstelzer:
dgdebug_tty.exe started out as an extremely basic stopgap to make dgt work on Windows. But it's actually taken surprisingly little work to turn it into a functional program in its own right. This branch is an experiment toward that end. If it can achieve basically the same functionality as dgdebug on Mac and Linux, then we can deprecate the (generally buggier and less-featureful) Windows Glk frontend.

Things the Windows TTY implementation still lacks:

- [x] Turn off input echoing, which is necessary for the history implementation—but it sounds like the shell can do that for us
- [x] Ctrl-C to interrupt an infinite loop (i.e. while processing is happening)
- [x] Ctrl-C to interrupt a flood of output (i.e. at the [more] prompt)
- [x] Terminal size measuring so explicit --width and --height don't have to be put in
- [x] Formatting with ANSI escapes
- [x] Compatibility with pipes for dgt usage

Once we've got all of these, I'll feel comfortable making it the official dgdebug.exe for the future.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
       - name: make
         run: make --directory=./src
       - name: make (windows)
-        run: make --directory=./src dgdebug.exe dgdebug_tty.exe dialogc.exe
+        run: make --directory=./src dgdebug.exe dgdebug_gui.exe dialogc.exe
       - name: make test
         run: make test
       - name: make test-6502
@@ -48,7 +48,7 @@ jobs:
           name: win32
           path: |
             src/dgdebug.exe
-            src/dgdebug_tty.exe
+            src/dgdebug_gui.exe
             src/dialogc.exe
 
   build-manual:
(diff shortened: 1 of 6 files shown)


dstelzer:
I believe terminal size measuring and both types of ctrl-C should now be handled. I'm not going to check them until they're tested, though.

dstelzer:
Compatibility with pipes is being investigated by hlship, so I'm going to call it Not My Problem until bug reports come in. Huzzah!

dstelzer:
It looks like the community [wants the command-line version to be standard](https://intfiction.org/t/dialog-wishlist/70278/510?u=draconis), so I've made that change too.

PR #247, opened 2026-04-27
Update license

dstelzer:
This PR adds an informal note to LICENSE.txt that clarifies what I believe to be Linus's intentions. [See here for discussion.](https://intfiction.org/t/licensing-the-dialog-standard-library/80032?u=draconis) Once we have confirmation from Linus, we can turn it into something more official, but for now hopefully people will accept a "we won't sue you for doing this".


dstelzer:
I got approval from everyone who's contributed to stdlib.dg, including Linus, so I've now properly updated license.txt.

PR #248, opened 2026-04-27
Add Unicode version of standard library

dstelzer:
Automatically generated by the Makefile, with curly quotes instead of straight ones.


dstelzer:
After some forum discussion, I will **not** be merging this PR. Keeping it open anyway until I can move it to a different repository.

PR #250, opened 2026-04-29
remove unneeded patch

susan-davis:
There was a patch in `unit.dg` for problems with `($ contains sublist $)` in 0m/03. It isn't needed any more, so get rid of it.

### DIFF
--- a/unit.dg
+++ b/unit.dg
@@ -105,11 +105,7 @@
 
 ($Test has failed)
 	(list of failures $List)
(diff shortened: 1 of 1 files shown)


dstelzer (approved):
Looks good to me! I'm curious what we did that fixed the sublist issues, though; what was the previous behavior?

susan-davis:
I don't know what fixed the problems, but I'm not able to reproduce them in 1.x, so.

PR #258, opened 2026-05-04
add copyright and licensing notices

susan-davis:
We're being explicit about licensing, so add notices to contributed files that match that.


dstelzer:
Good catch!

PR #264, opened 2026-05-09
redesign unit.dg to eliminate (list of tests $)

susan-davis:
`unit.dg` now expects each test to be flagged with trait `(test $)` and have a predicate `(run $)` that executes the test and succeeds or fails. The `(list of tests $)` global variable is gone.

### DIFF
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -1428,7 +1428,7 @@ int debugger(int argc, char **argv) {
 			case 'T':
 				io_tag_lines = 1;
 				break;
-			case 'u': // --no-warn-not-topic --quit --height=-1
+			case 'u': // --quit --no-header --height=-1
 				quitopt = 1;
 				force_height = -1;
 				suppress_header = 1;
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -1332,8 +1332,7 @@ void usage(char *prgname) {
 	fprintf(stderr, "--numbered  -N      Show call depth with numbers during tracing.\n");
 	fprintf(stderr, "--tag-lines -T      Prepend output with \"  \", input with \"> \" or \") \".\n");
 	fprintf(stderr, "--no-header         Don't show version information at startup.\n");
-	fprintf(stderr, "--unit-test -u      Same as --no-warn-not-topic --quit --height=-1\n");
-	fprintf(stderr, "                    --no-header.\n");
+	fprintf(stderr, "--unit-test -u      Same as --quit --height=-1 --no-header.\n");
 }
 
 extern int topic_warning_level; // Defined in frontend.c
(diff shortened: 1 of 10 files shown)


dstelzer:
Looks good to me! Question, though: should `--no-warn-not-topic` be removed from `--unit-test` now?

susan-davis:
Done. Also, we should do a 1b/02 release, to keep folks from writing tests around the `(list of tests $)` paradigm.

dstelzer:
Looks good to me! My only remaining concern isn't about this change but about the time system overall: `(suppress this tick)` and `(inhibit next tick)` sound very similar. Is there a reason to keep them separate?

susan-davis:
`(suspend this tick)` prevents more time from being counted when the next tick happens, but the tick still does happen, with all the `(every tick)` actions hanging on it. `(inhibit next tick)` prevents a tick from happening at all.

Honestly, I had forgotten the existence of `(inhibit next tick)` when I had a need for "free" actions. Are there cases where we want all of the next-tick things to happen more frequently than one minute? Maybe not.

`(time.dg)` and `(utils.dg)` were just thrown in to provide examples for writing unit tests, but if we're reviewing them as actual library code, that's also cool. I have a `https://github.com/susan-davis/dialog-extensions` repository where things that might be useful to more than just my WIP get stashed; do we want a `https://github.com/dialog-if/dialog-extensions` directory for things that don't make the cut to wind up in `stdlib.dg`?

dstelzer:
Ah, if they're meant as demos rather than extensions, then don't worry about this! I would personally keep time and ticks in sync, so that e.g. an every tick rule that moves a guard around happens at predictable time intervals, but that only matters if we expect people to use this code verbatim.

dstelzer (approved):

susan-davis:
No, they're meant as extensions, but we haven't sorted out where those live yet, so they're in the `test/unit` directory as example code. (`utils.dg` includes mostly things that are broadly useful, but one or two which are specific to my own needs.) Should `time.dg` live somewhere where folks can find it for actual use?

I think I'm going to YAGNI `(suspend this tick)`.

dstelzer:
I think I asked on the forum before if people want extensions to live in this repository in a `/lib` folder, or in a separate repository (like Inform extensions). I don't remember if we came to any consensus, though.

Personally, I think I'd lean toward a separate repository, so that extension updates aren't bound to Dialog's release cadence. But I should dig up that forum thread and see what people said there.

susan-davis:
Dixitque Dea fiat dialog-extensions, et facta est [dialog-extensions](https://github.com/Dialog-IF/dialog-extensions).

dstelzer:
Looks good to me! We might want to put them into subdirectories as we accumulate more, but that's an issue for that repository (and for the future), not this one in the present.

PR #265, opened 2026-05-10
Tabs to spaces

susan-davis:
Change tabs to spaces in the license text at the bottom of the Dialog libraries (`stdlib.dg`, `stddebug.dg`, and `unit.dg`) so that it fits in 80 columns.


susan-davis:
Branched from wrong branch. Will redo.

PR #266, opened 2026-05-10
make license text fit in 80 columns

susan-davis:
Convert tabs to spaces to get the license text at the bottom of the libraries (`stdlib.dg`, `stddebug.dg`, and `unit.dg`) to fit in 80 columns and not look like ass.


dstelzer (approved):
Looks good to me! If this looks good on your screen, then go ahead and merge!

PR #274, opened 2026-05-11
Make dgdebug respond to Ctrl-D at [more] prompt

### LINKED ISSUE #267: Ctrl-D at a [more] prompt does not end the debugger
The debugger looks for Ctrl-C at the [more] prompts, but does not recognize Ctrl-D (EOF). It would be nice if that were consistent—especially since the `--height` option can lead to [more] prompts when reading from a file now.

dstelzer:
Fixes #267

### DIFF
--- a/src/term_tty.c
+++ b/src/term_tty.c
@@ -97,7 +97,7 @@ void morefunc() {
 	fflush(stdout);
 	i = term_getkey("[more]");
 	printf("\r      \r");
-	if(i == -1) {
+	if(i == -1 || i == 4) { // EOF
 		exit(0);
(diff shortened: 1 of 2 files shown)


src/term_tty.c
```diff
@@ -97,7 +97,7 @@ void morefunc() {
 	fflush(stdout);
 	i = term_getkey("[more]");
 	printf("\r      \r");
-	if(i == -1) {
+	if(i == -1 || i == 4) { // EOF
```
dstelzer:
```suggestion
	if(i == -1 || i == 4) { // EOF, Ctrl-D
```

PR #278, opened 2026-05-12
Use aamrun binary if available

### LINKED ISSUE #174: Include Homebrew install of aamachine in search list for aamrun.py
The build now fails if you don't also have `aamachine` cloned in the same workspace directory. To decouple the two projects, allow `aamrun.py` to get `nodefrontend.js` from the `aamachine` installation in Homebrew. (This will require the Homebrew formula for `aamachine` to include `nodefrontend.js`.)

(While we're at it, changing aamrun.py to aamrun.sh wouldn't be out of place, with a tiny bit of shell hacking.)

dstelzer:
Fixes #174 , assuming Homebrew puts aamrun on the $PATH.

Which I'm not sure it does. I'd appreciate if either @susan-davis or @hlship can test this with a proper Homebrew installation on Mac and make sure the Python script is locating it correctly. Currently it runs `which aamrun`, and if that succeeds, it assumes it can use `aamrun -s 1234 file.aastory` instead of `node path/to/nodefrontend.js -s 1234 file.aastory`.

### DIFF
--- a/bin/aamrun.py
+++ b/bin/aamrun.py
@@ -5,6 +5,7 @@
 from pathlib import Path
 from sys import argv, exit, stderr
 import subprocess as sp
+import shutil
 
 USAGE = f'''Usage: {argv[0]} [file.aastory]
 Attempts to play an .aastory file in the Node interpreter (with consistent random seed for testing), despite varying install locations.'''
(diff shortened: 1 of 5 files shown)


susan-davis:
`aamrun` is in /opt/homebrew/bin, right where it should be.

dstelzer:
And does the new Python script find and execute it successfully? It does on my machine, but I specifically want to make sure this works on Mac before merging, if possible.

susan-davis:
It does.

dstelzer:
Excellent!

PR #281, opened 2026-05-13
Write more of library chapter 12

susan-davis:
There's more to say about testing, so say it.


susan-davis:
I have more to write, but let's get the next instalment in.

dstelzer:
This all looks great! I'd say add a note in readme.txt so that people will know to look for the new manual chapter, then go ahead and merge!

dstelzer (approved):

PR #285, opened 2026-05-14
bump versions

susan-davis:
Post-release bump of versions.

### DIFF
--- a/manual/antora.yml
+++ b/manual/antora.yml
@@ -1,14 +1,14 @@
 name: dialog
 title: Dialog
-version: 1b02
+version: 1c01-dev
 nav:
   - modules/ROOT/nav.adoc
   - modules/lang/nav.adoc
   - modules/lib/nav.adoc
 asciidoc:
   attributes:
-    lang-ver: 1b
-    lib-ver: 1.2.1
+    lang-ver: 1c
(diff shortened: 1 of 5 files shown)


dstelzer:
I don't think we need to bump the library version to 1.3.0 yet—if the library doesn't include any 1c/01-exclusive features, we should keep it at 1.2.x so people know they can still use the new library with an older compiler.

(The reason it needed to be bumped to 1.2 with 1b/01 is because we added `(transcript active)` to some actions.)

susan-davis:
Okay, so we're not keeping the library and compiler versions in lockstep, then? If I have library version X, how do I know what version of the compiler I need?

dstelzer:
The way I've been doing it, library version X.Y.Z needs compiler version X.Y or better. So a 1.1.* library needs 1a/01 or later, a 1.2.* library needs 1b/01, a 1.3.* library needs 1c/01, and so on.

The intent is that people can choose to update their library without updating their compiler. But honestly I'm not sure if that's a common issue; if we want the library and compiler versions to be in lockstep again, I wouldn't be too opposed. (We'll just have to adjust all the places where the version numbers are used, like the manual and the Github actions.)

PR #309, opened 2026-06-06
report fatal error on exit

susan-davis:
`unit.dg` doesn't currently tell the user what the error is when it exits with a fatal error. Use `(report fatal error $)` to fix this.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -48,6 +48,9 @@ Release notes:
 		Compiler: objects produced by (generate $ $) no longer have
 		author-accessible names, which could confuse the compiler.
 
+		Unit test runner: unit.dg now tells the author what the
+		fatal error was when it exits on a fatal error.
+
 	1c/01, Lib 1.2.2:
 
 		Library: fixed a bug where ASK FOR SOMETHING queried a random
--- a/unit.dg
+++ b/unit.dg
@@ -129,7 +129,8 @@
 	(if) (using color) (then)
 		 (body style @red-bar)
     (endif)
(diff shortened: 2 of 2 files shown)


dstelzer:
This looks good, but is unit.dg always expected to be used with stdlib.dg? If not, I would recommend including this block:
```
(report fatal error 1)	Heap space exhausted.
(report fatal error 2)	Auxiliary heap space exhausted.
(report fatal error 3)	Type error: Expected object.
(report fatal error 4)	Type error: Expected bound value.
(report fatal error 5)	Invalid dynamic operation.
(report fatal error 6)	Long-term heap space exhausted.
(report fatal error 7)	Invalid output state.
(report fatal error $)	Invalid error code!
```

A redundant definition won't hurt anything, and if this predicate failed, then the `(quit 2)` wouldn't run and the debugger would exit with status 0.

I also usually put `(display memory statistics)` in an error entry point, but I'm not sure that would be useful in this situation. You'll already know which heap was exhausted.

dstelzer:
This is also a good place to decide: do you want the version number in unit.dg to always match the one in stdlib.dg, even if the libraries are updated independently? If so, it's easy to adjust bin/bump-version.sh to ensure it; if not, easier to do manually.

susan-davis:
I've stuck with the `stdlib.dg` version by default, just because it's the one that goes with a particular release, but that doesn't have to be the case.

I had assumed that I'd used `stdlib.dg` predicates in `unit.dg`, but looking again, it looks like they're just builtins (except for `(report error code $)`, so no need to couple the two.

PR #312, opened 2026-06-07
Add command-line option to disable extended ZSCII

### LINKED ISSUE #311: Add option to disable extended ZSCII
Some retro interpreters can handle check_unicode but not custom extended ZSCII. A command-line option to avoid using extended ZSCII will improve compatibility (though it'll make parsing worse on modern terps).

dstelzer:
This leads to larger files and worse parsing, but improves compatibility with retro interpreters that don't support Unicode translation tables. It shouldn't be needed often, but when it is, it's good to have it.

Also used this opportunity to clean up how configuration variables are passed to backends.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -48,8 +48,12 @@ Release notes:
 		Compiler: objects produced by (generate $ $) no longer have
 		author-accessible names, which could confuse the compiler.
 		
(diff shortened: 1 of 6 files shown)


dstelzer:
Fixes #312

PR #315, opened 2026-06-10
Update test cases for new Å bugfix

dstelzer:
The Å-machine previously had a bug involving paragraph breaks after divs, and our test cases were written to accommodate that. Now the bug is fixed. This PR updates the test cases to no longer expect the old (bad) behavior.


dstelzer:
Note: requires https://github.com/Dialog-IF/aamachine/pull/76

PR #316, opened 2026-06-14
Continue writing chapter 12

susan-davis:
Continue with chapter 12, covering

- testing the object model
- using `regtest.py`
- using the Skein
- using `dfrotz`, `aamrun`, and `diff`
- using `make`
- multiplatform testing
- using `dgt` as a `make` substitute
- troubleshooting failing tests

### DIFF
--- a/bin/bump-version.sh
+++ b/bin/bump-version.sh
@@ -14,7 +14,6 @@ ANTORA_VERSION=$(echo "$COMPILER_VERSION" | tr -d /)
 
 sed -Ei "s|DIALOG_VERSION=.+|DIALOG_VERSION=$COMPILER_VERSION|" src/Makefile
 sed -Ei "s|\(library version\) Library version .+|(library version) Library version $LIBRARY_VERSION.|" stdlib.dg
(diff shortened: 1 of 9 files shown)


susan-davis:
Left a couple of bits out of the "unit testing your game" section, because of the problem with `(collect words) . . . (into $)`; will add those after a fix is merged.

susan-davis:
What the hell?! Text that I wrote and committed isn't in the branch. Fixing.

dstelzer:
Looks good to me! Instead of `(collect words) (do something) (into $)`, I would recommend:

```
(style class @invisible)
	display: none;

... (div @invisible) (do something) ...
```

This is the preferred way to keep output from being sent to the screen, since `(collect words)` has heavy implications for the optimizer.

dstelzer (approved):

susan-davis:
> This is the preferred way to keep output from being sent to the screen, since `(collect words)` has heavy implications for the optimizer.

`(collect words` isn't about stopping the output from going to the screen, it's about testing _what_ the output was. But in many cases, that's better done with an all-up test in the interpreter (`diff` or `regtest.py` or the Skein) than in a unit test.

dstelzer:
Ahh, I see! Yes, in that case that will unfortunately have to wait for the bug to be fixed.

PR #323, opened 2026-06-16
Write all-up testing section of chapter 12

susan-davis:
We've finished unit tests, now talk about interpreter testing with `diff`, `regtest.py`, and the `dgt` Skein.


manual/modules/lib/pages/testing.adoc
```diff
@@ -1096,6 +1117,62 @@ rather than leaving them in `(run *)`: the test is testing that it's
 the _second_ time that you do something foolish in the dark that wipes
 out the message, not the first time when you're warned.
 
+However, when we run this, it fails! The release 2 version of _Cloak
+of Darkness_ implements the "don't screw around in the darkness"
+warnings using `(select) . . . (or) . . . (stopping)`, which doesn't
+reset between tests. (As of lDialog does not have a mechanism for resetting a
```
dstelzer: Typo?

manual/modules/lib/pages/testing.adoc
```diff
@@ -1180,7 +1257,592 @@ Finally, let's test the defeat condition:
...
+== All-Up Tests in the Interpreter
+
+Unit testing tests small units of your code, and validates that it's
+doing what you expected it to do, and that you didn't break anything
+by making changes. Unit tests are small, quick to write, and lightning
+fast to execute. On modern hardware, you can execute hundreds or
+thousands of them in a second (up to 16383 per test file), and get
+answers very quickly.
+
+End-to-end, or all-up, tests, sometimes called _integration tests_,
```
dstelzer: "End-to-end, all-up, or integration tests"?

manual/modules/lib/pages/testing.adoc
```diff
@@ -1180,7 +1257,592 @@ Finally, let's test the defeat condition:
...
+>
+----
+
+Having suppressed the banner, we can create an output file with
+`dgdebug -u cloak.dg no-banner.dg stdlib.dg < cloak.dg >
+cloak.out`. That will give us the output we saw earlier, minus the
+banner, in a file that we can later test against. Since we've reviewed
+it and seen that the output is exactly what we expect, we can "bless"
+it as being the correct output that we expect by copying it to another
+file -- we'll call it `cloak.gold` -- which we can compare against any
```
dstelzer, line 1475: Explain that it's .gold because it's the "gold standard" everything will be compared against?

manual/modules/lib/pages/testing.adoc
```diff
@@ -1180,7 +1257,592 @@ Finally, let's test the defeat condition:
...
+that emulates a Linux shell and its associated commands.
+
+[#branches]
+==== Testing Multiple Branches
+
+Remarkably, for such a small game, _Cloak of Darkness_ is a game with
+multiple endings: you can preserve the message in the sawdust and win,
+or you can accidentally destroy it, and lose. The two endings are
+mutually exclusive, so there's no single walkthrough that can cover
+them both.
```
dstelzer: I frequently use UNDO in my integration tests to cover diverging branches that are only a move or two long (e.g. an instant death option). That might be worth mentioning here as well, though of course for longer branches it's better to have multiple tests.

manual/modules/lib/pages/testing.adoc
```diff
@@ -1180,7 +1257,592 @@ Finally, let's test the defeat condition:
...
+
+One of Dialog's strengths is that it can produce games playable on
+multiple platforms: on a Z-machine interpreter, in a web browser, or
+on vintage computers such as the Commodore 64. (The C64 is the only
+supported vintage platform as of Dialog 1c/02 and Å-machine 1.0.2, but
+others may be added later.) In principle, a Dialog game should run
+equally well on all platforms, so long as it fits within the resource
+limitations of the Z-machine or the Commodore 64. In practice, testing
+on each of the platforms on which you intend to release your game is a
+good practice.
```
dstelzer: Since the word "practice" was used at the start of this sentence, I'd switch to a synonym, like "is a good habit to get into" or "is a good idea".

manual/modules/lib/pages/testing.adoc
```diff
@@ -1180,7 +1257,592 @@ Finally, let's test the defeat condition:
...
+$ echofrotz.py -m -q cloak-test.z5 <win.in >win-zmachine.out
+$ diff win-zmachine.out win-zmachine.gold
+$ echofrotz.py -m -q cloak-test.z5 <lose.in >lose-zmachine.out
+$ diff lose-zmachine.out lose-zmachine.gold
+----
+
+We'll need a copy of the Å-machine distribution in order to test the
+Å-machine version of our game on both the web and the Commodore 64.
+See the xref:ROOT:software.adoc[Software Page] for where you can find it.
+Specifically, you'll need the `aamrun` and `aambox` utilities from it,
```
dstelzer, line 1621: Do we currently distribute aambox as an executable? I think people will need to build it from source (and also its frontend).

manual/modules/lib/pages/testing.adoc
```diff
@@ -1180,7 +1257,592 @@ Finally, let's test the defeat condition:
...
+
+...and with that, we've thoroughly tested our game, with both unit and
+in-browser tests.
+
+[#regtest]
+=== Using `regtest.py`
+
+Another alternative for doing all-up tests is the `regtest.py` script,
+by Andrew Plotkin. Like the other Python scripts that we've been
+using, it can be found in the `bin` directory in the Dialog source
```
dstelzer, line 1650: Probably worth being specific here: it's included in the Dialog _repository_, but not in the releases, even though those releases include the source of the compiler.

test/cloak/no-banner.dg
```diff
@@ -0,0 +1,3 @@
+%% no-banner.dg
+
+(banner)
```
dstelzer, line 3: This is a fine way to do it, but I tend to make my banner replacement print \<BANNER> or the like, to make sure it's actually getting called.

test/cloak/win-debugger.gold
```diff
@@ -0,0 +1,55 @@
+
+
+Hurrying through the rainswept November night, you're glad to see the bright
+lights of the Opera House. It's surprising that there aren't more people about
+but, hey, what do you expect in a cheap demo game...?
+
+Foyer of the Opera House
+<[me] You> are standing in a spacious hall, splendidly decorated in red and
```
dstelzer, line 8: These links are one prominent reason why the output can differ between platforms. That might be worth calling out in the explanation.

dstelzer (dismissed):
Looks good! I flagged some minor things but no major ones.

manual/modules/lib/pages/testing.adoc
susan-davis: Do we want to drop those four Python scripts into the Dialog distro?

susan-davis:
It's saying there's one requested change, but it's not showing me what the change is, and all conversations are resolved.

PR #325, opened 2026-06-22
write the "make" section of chapter 12

susan-davis:


dstelzer:
This looks fine, though I'm not sure if it's the Dialog manual's responsibility to teach people how to use `make`. If you think it should be included, though, it does no harm.

dstelzer (approved):

PR #331, opened 2026-08-14
Build universal (arm64+x86_64) macOS binaries

### LINKED ISSUE #330: Support Mac Intel as well as Mac Arm
When compiling, the binaries should be compiled to support Intel macs as well as ARM.  AFAIK there's a way to "fuse" the two into a single binary.

hlship:
## Summary
- Build dialogc/dgdebug as fat binaries on macOS by passing `-arch arm64 -arch x86_64` in CFLAGS/LDFLAGS when `uname -s` is Darwin, leaving Linux and the mingw cross-builds untouched.
- Rename the macOS CI artifact from `macos-arm64` to `macos-universal` to reflect the change (both the upload step and the release-bundling step).

Closes #330

## Test plan
- [x] `make --directory=./src clean all` on macOS (Apple Silicon) and confirmed `lipo -info` reports both `x86_64` and `arm64` for `dialogc` and `dgdebug`
- [x] CI: `build-macos` job builds and `make test` passes with the universal binaries

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,7 +88,7 @@ jobs:
         run: make test
       - uses: actions/upload-artifact@v4
         with:
-          name: macos-arm64
+          name: macos-universal
           path: |
             src/dgdebug
             src/dialogc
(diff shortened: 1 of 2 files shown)


dstelzer:
@hlship Is this PR ready to merge? It's not marked as a draft, and the Makefile updates look good to me (as a non-Mac-user), but one of the checkboxes is still clear.

hlship:
The CI steps worked properly. I've built locally and tested the binaries on both platforms. I think we're good to go, but I did just push a docs update.

hlship:
I screwed something up Git wise and it was easier to open a new PR.

Issue #291, opened 2026-05-19
Spans inside closures create hidden variables


dstelzer:
There are certain pieces of syntax that create their own variables for internal use. In particular, anything that discards choice points (spans, divs, conditionals, at most once) needs a variable of its own to function.

But if these pieces of syntax appear inside a closure, those hidden variables will be considered parameters to the closure, which means a closure will be called with unbound parameters. This means _no_ closure can assume its parameters are bound, which leads to a flood of warnings and optimization issues.

(For example, if any closure contains `(the $Obj)`, it needs to know all its parameters will be bound, including `$Obj`.)

These hidden variables can't be accessed by anything except their own bits of syntax, so they should not be considered parameters to the closure. They can be recognized by a star at the front of their name.

PR #347, opened 2026-08-29
Prevent hidden variables from interfering with closures

dstelzer:
Every div, span, and certain other constructions creates a hidden variable that's used for internal purposes.

These hidden variables should not be considered when determining if the inputs to a closure are bound or not.

For example, this source:

```
(style class @red) color: red;
(style class @green) color: green;

(interface (this should be bound $<This))
(this should be bound $This) $This

(program entry point)
	(query { (span @red) Red! })
	(query { (span @green) Green! })
	($B = 123)
	(query { (this should be bound $B) })
```

With the current compiler, it produces this:
```
Warning: closure-span.dg, line 11: Parameter #1 of (this should be bound $) can be (partially) unbound, which violates the interface declaration at closure-span.dg:4.
Warning: closure-span.dg, line 11: Parameter #1 of (this should be bound $) can be (partially) unbound, which violates the interface declaration at closure-span.dg:4.
```

Because the red and green spans introduce hidden variables that mean `( invoke-closure $ $ $)` does not have fully-bound parameters.

With this PR, those hidden variables are not included in the signature of `( invoke-closure $ $ $)` and thus will not be considered by the optimizer. Thus, no warnings.

This was previously reported as #291 and reported as fixed in #292 . However, that PR only fixed it for the debugger. This PR fixes it for the compiler as well.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -49,9 +49,12 @@ Release notes:
 		The problem isn't really fixed, but it will no longer corrupt
 		your game text.
 
-		Unit test runner: (game over) and (game over $) are
-		now overridden in unit.dg, to enable testing of win
-		conditions.
+		Compiler: re-fixed the obscure bug with divs and spans inside
+		closures, which was previously only fixed for the debugger, not
+		for the compiler. Oops!
+
+		Unit test runner: (game over) and (game over $) are now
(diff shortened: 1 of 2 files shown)


dstelzer:
Fixes #291 properly this time

PR #349, opened 2026-08-30
Fail on multiplication overflow

### LINKED ISSUE #328: Multiplication overflow should make the predicate fail
See [here](https://intfiction.org/t/an-odd-bit-of-undefined-behavior/80914?u=draconis) for discussion. It's odd that `($ times $ into $)` is the only math predicate that returns an incorrect result instead of failing when a calculation overflows. It would be nice to fix that.

dstelzer:
Fixes #328 .

At least, this fixes it on Z-machine and debugger. The Å-machine will need a spec update.

### DIFF
--- a/manual/modules/lang/pages/builtins.adoc
+++ b/manual/modules/lang/pages/builtins.adoc
@@ -73,8 +73,10 @@ fails.
 
 `A` and `B` must be bound to numbers; `C` is unified with
 their product. If the product is outside the valid range of numbers, the query
-succeeds, but the numeric result is unpredictable (i.e. it depends on the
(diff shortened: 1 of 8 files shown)


dstelzer:
Marking this unready to merge until the Å-machine version is ready.

dstelzer:
The Å-machine update will use the same opcode, so Dialog doesn't necessarily need to be updated, but we should bump the Å-machine spec version to ensure the proper behavior.

dstelzer:
Tests will now fail until the new Å-machine version is live. I'll leave this PR as a draft until then.

PR #352, opened 2026-09-05
feat: combined Intel + ARM binaries for OS X

hlship:
This replaces the prior PR, which I'll close.


dstelzer:
Looks good to me! I have no way of testing this myself, but the test suite runs on Mac, so it's certainly not a downgrade at least.

On behavior vs behaviour in the manuals, we should open a new discussion about that. I'm pretty sure we currently have a mix of American and British spellings.

But otherwise, all looks good! Merge when ready.

PR #353, opened 2026-09-09
Added WASI target, fixed platform-dependent qsort

sehugg:
This enables a WebAssembly build via WASI SDK (https://github.com/webassembly/wasi-sdk)

The WASI port also exposed a property of qsort() that behaves differently on different platforms, now dialogc should generate the same .aastory bytes on all platforms.


dstelzer:
The changes look good! Do we want to distribute this build along with the others? Or just leave it for people who are building from source?

sehugg:
Wouldn't hurt, I can add it to the Github workflow.

sehugg:
I forgot dgdebug ... it's included now, but like the Windows version, won't have nifty terminal features.

dstelzer:
All looks good to me then. I'm a bit leery of hardcoding a specific version of the SDK in the Github build actions, but if the SDK changes a lot between versions, then I suppose it's better than having everything suddenly break.

dstelzer:
Last request: add a line to the version history in the README mentioning the new WASM build?

PR #357, opened 2026-09-11
Improve locality of strings in WRIT chunk

sehugg:
The `analyze_strings()` function sorts all strings by ascending size, so they can be placed in short/medium/long tiers for encoding. This scatters related strings uniformly across the WRIT chunk, which means the 8-bit platforms have to read more pages. One good example is the intro text, which might be scattered across 5 different pages.

The fix is to still do the initial sort, but then we re-sort the medium and long tiers back into their original index order. Thus every string stays in the same tier, but locality is improved. This reduces WRIT chunk reads by about 30 percent on 6502 platforms.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -48,6 +48,11 @@ Project website:
 
 Release notes:
 
+	1c/04, Lib 1.2.4:
+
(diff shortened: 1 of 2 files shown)


dstelzer:
Brilliant!

